PR #1, opened 2025-11-04
Updated pseudo-ish code

nikil-ravi:


nikil-ravi:
To run: I did:

```
 cd /Users/nikilravi/Desktop/Projects/vals/agentic-harness && PYTHONPATH=./src uv run python -m main --agent math_agent --benchmark addition 2>&1 | tail -15
 ```

File: pyproject.toml [hunk 1]
```diff
@@ -6,6 +6,7 @@ readme = "README.md"
 requires-python = ">=3.13"
 
 dependencies = [
+  "vals-model-library @ git+https://github.com/vals-ai/model-library.git",
```
JarettForzano, line 9: This dependency is outdated, can you make it model-library?

agents/math_agent/main.py
```diff
@@ -0,0 +1,40 @@
...
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), "../../src"))
+
+from agent_runner import AgentRunner
+
+
+class MathAgent(AgentRunner):
+    """Solves addition problems."""
+
+    def run(self, input_items: list) -> dict[str, str]:
```
JarettForzano: Can you type the list? `list[InputItem]`

benchmarks/addition/main.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def evaluate_output(output: dict[str, str], run_id: str) -> dict:
+    """Evaluate agent answers against expected results."""
+    dataset = get_dataset()
+    correct = 0
+    total = len(dataset)
+
+    for task_id, task in dataset.items():
+        if task_id in output:
```
JarettForzano: I would try to find an easier way to do this and expose the task_id in the interface. When evaluating we should know that the output provided belongs to a specific task. But lmk if im misinterpreting what this is doing

agents/math_agent/main.py
```diff
@@ -0,0 +1,40 @@
+"""Simple math agent that solves addition problems."""
+
+import os
+import sys
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), "../../src"))
```
JarettForzano: Whats the reasoning behind needing to add an extra path?

src/benchmark_runner.py
```diff
@@ -0,0 +1,122 @@
...
+        if not benchmark_path.exists():
+            raise FileNotFoundError(f"Benchmark not found: {benchmark_name} at {benchmark_path}")
+
+        benchmark_module = self._load_module(benchmark_path)
+
+        class Benchmark:
+            def __init__(self, module: ModuleType, name: str):
+                self.module = module
+                self.name = name
+                self.requires_sandbox = getattr(module, "requires_sandbox", False)
```
JarettForzano: instead of sandbox can we go with docker? I think it would be easier to grasp

pyproject.toml
nikil-ravi: Hmm is it not vals-model-library? https://github.com/vals-ai/model-library/blob/17bffb889ffc3cce71b09861062962feac25d120/pyproject.toml#L2

agents/math_agent/main.py
nikil-ravi: I made src/ a package now if that's what you meant

benchmarks/addition/main.py
nikil-ravi: Made it a list

JarettForzano (approved):
LGTM

PR #3, opened 2025-11-19
Implement FAB

oalmatov:
# Changes
- Reorganized file structure to stay consistent with model library
- Deleted old classes
- Copied over `Agent` from FAB and wrapped it with the agentic harness' `Agent` class
- Implemented `FinanceAgentBenchmark` class
- Added `registry.py` to easily load agents and benchmarks
- implemented very basic runner function

# Next steps
- How will we store configs?
  - We should be able to swap models / other params inside agents (i.e. Claude Code with sonnet 4.5, haiku 4.5, etc.)
- Runners
  - containerization
  - running agents in a distributed system
  - also a benchmark runner that schedules benchmarks, although idk if this should live in the agentic harness vs benchmarks repo


agentic_harness/registry.py
```diff
@@ -0,0 +1,100 @@
...
+    return decorator
+
+def register_benchmark(name: str) -> Callable[[Type[BenchmarkT]], Type[BenchmarkT]]:
+    """Decorator that registers a Benchmark subclass under ``name``."""
+
+    def decorator(cls: Type[BenchmarkT]) -> Type[BenchmarkT]:
+        _benchmark_registry[name] = cls
+        return cls
+
+    return decorator
```
JarettForzano: What is the reasoning behind lazy loading all of the agents and benchmark modules? At run time would we ever be needing more than one agent or benchmark? If not I don't think this design pattern may be applicable here,
Inside of the model library it makes sense since
1. We have some model classes we may not want to expose
2. We want all of the classes available at run time and instantiated -> since we are making more than one request

But for here when we start the process we would be loading at most one benchmark and one agent, at runtime we would specify the module like so
`--benchmark fab`
and then we could just do something like
```
if not importlib.util.find_spec(benchmark):
    raise ValueError(f"Benchmark {benchmark} not found")


module = importlib.import_module(benchmark)
parsed_cls = cast(type[Benchmark, getattr(module, class_name))

parser = parsed_cls(
  field_1=....,
  field_2=....
)
```

which would simplify a decent chunk of abstraction and setup here

agentic_harness/base/agent.py
```diff
@@ -0,0 +1,11 @@
+from abc import ABC, abstractmethod
+
+from model_library.base import InputItem, QueryResult
+
+
+class Agent(ABC):
+    """Interface for models that execute a single task and return a result."""
+
+    @abstractmethod
+    async def run(self, input_items: list[InputItem]) -> QueryResult:
```
JarettForzano, line 10: I think when we initially talked i had mentioned using the query result specifically for the agent class but later re-evaluated the return type. 

For some benchmarks we are going to want the returned artifact from the agent to be different. Its going to be hard to standardize that, so I think there are two options
1. We make a tuple `[QueryResult, dict[str, Any]]`
2. We just keep it a dict

If we keep it a tuple then that gives us flexibility since we can enforce a type for what we want to evaluate while providing a means to export additional information. the dict that is returned would just be dumped into the exported file ontop of whatever else we want to export from the evaluation- you can maybe just think of it as metadata?

oalmatov: gonna keep it in QueryResult.raw

agentic_harness/registry.py
oalmatov: ur right removed

JarettForzano (approved):

PR #8, opened 2025-12-12
WIP: Daytona

JarettForzano:


src/base/types.py
JarettForzano, line 1: Not to self clean up AgentConfig

src/benchmarks/base_benchmark.py
JarettForzano, line 1: Removed ABC and used as base class, migrated to benchmarks/

src/datasets/test/dataset.py
```diff
@@ -0,0 +1,36 @@
...
+        return Path("submodules/claude_code/Dockerfile.external")
+
+    @override
+    async def create(self) -> list[TaskGroup]:
+        task_groups: list[TaskGroup] = []
+        with open(self.specsheet_path, "r") as f:
+            specsheet = f.read()
+
+        if not self.dockerfile_path.exists():
+            raise ValueError(f"Dockerfile does not exist at {self.dockerfile_path}")
```
JarettForzano: Need cleaner way to add dockerfiles inside of dataset

src/entry/__main__.py
JarettForzano, line 1: Need top level docstring

src/registry.py
JarettForzano, line 1: Clean up parsing in here

PR #14, opened 2026-01-08
First pass at tracker service

oalmatov:
# Tracker Service Implementation

Implemented the tracker service for orchestrating benchmark runs with agent contracts.

## tracker service
  - `main.py`: web server for uploading contract and starting an agent
  - `sandbox.py`: helper functions for interacting with daytona sandbox
  - `s3.py`: utilities for uploading to s3
  - integration tests for daytona sandbox.

## general
- basic CLI for starting a run
- reorganized contracts folder
- deleted unused files

## Running Tests

```bash
# All tracker tests
make tracker-test
```

## Development

```bash
# Install dependencies
make tracker-install
make swebench-install

# Start service (you may need to change the ports in the Makefile)
make tracker-dev
make swebench-dev

## Usage

```bash
# Run a benchmark task
harness run --contract contracts/claude_code/ --benchmark swebench
```

**CLI is VERY raw and a work in progress.**


contracts/claude_code/setup.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+# Setup script for claude_code agent
+# Installs Claude Code CLI and sets up the environment
+
+set -euo pipefail
+
+# echo "Installing dependencies..."
+# apt update && apt install -y curl vim
```
oalmatov, line 8: Saw these in the claude code dockerfile that you had, but i believe these are not needed

services/tracker/src/tracker/benchmark_service.py
oalmatov, line 1: This is copied straight from what @JarettForzano gave me. Main thing I changed was add logs

services/tracker/src/tracker/sandbox.py
```diff
@@ -0,0 +1,233 @@
...
+                        dirs_to_create.add(dir_path)
+
+    # Create all necessary directories first
+    if dirs_to_create:
+        mkdir_cmd = "mkdir -p " + " ".join(sorted(dirs_to_create))
+        result = await sandbox.process.exec(mkdir_cmd)
+        if result.exit_code != 0:
+            raise SandboxError(f"Failed to create directories: {result.result}")
+
+    await sandbox.fs.upload_files(files_to_upload)
```
oalmatov, line 151: pretty complicated way to upload dirs. i hope they add a sandbox.fs method to just do this in one line

services/tracker/src/tracker/sandbox.py
```diff
@@ -0,0 +1,230 @@
...
+        raise SandboxError("Session command didn't return a command ID")
+
+    command = await sandbox.process.get_session_command(session_id, command_id)
+
+    timeout = 60
+    polling_interval = 5
+    while command.exit_code is None and timeout > 0:
+        await asyncio.sleep(polling_interval)
+        timeout -= polling_interval
+        command = await sandbox.process.get_session_command(session_id, command_id)
```
oalmatov, line 255: keep track of inflight sessions in db or redis.

if client runs with `--watch` we will poll an endpoint to check if the command has exited.

oalmatov: we should run the command and return instead of waiting for it to complete

services/tracker/tests/integration/test_sandbox.py
```diff
@@ -0,0 +1,179 @@
...
+        await upload_contract_to_sandbox(test_sandbox, "nested")
+
+        # Verify nested files exist
+        result = await test_sandbox.process.exec("test -f /bundle/contracts/nested/subdir/deep/file.py")
+        assert result.exit_code == 0, "Deeply nested file should exist"
+
+        result = await test_sandbox.process.exec("test -f /bundle/contracts/nested/other/file.txt")
+        assert result.exit_code == 0, "Other nested file should exist"
+
+    async def test_install_dependencies_errors_on_missing_bundle(
```
oalmatov, line 143: these tests aren't super comprehensive, because i need to create an actual contract bundle.

gonna do this in a separate PR when i actually add agentic harness as a dependency

src/agentic_harness/cli/utils.py
oalmatov, line 1: this is a mess. if you're not planning to add to this. i would prefer to just merge in the changes, and i'll have a follow up PR. if you do need to add stuff here, lmk and i'll refactor this right now so that merge conflicts aren't a pain

oalmatov: general idea: 

create a "bundle" with minimal dependencies. copy over only necessary files, zip up, and save them to s3.

oalmatov: if you have an idea for a better way to do this than constructing a `MINIMAL_PYPROJECT` let me know

Makefile
JarettForzano, line 1: 1. services should be ran through docker (no uv install service) ps its also why i left that command out of the README.md file inside of the service
2. all service related make commands in this file are redundant with the make file inside of the service
3. no need to expose service related unit tests inside of the agent harness make file considering that its an isolated service and no change to the agent harness will affect if the unit tests pass inside of the service

One of the reasons you are having SSL issues is probably because you are running the service locally.

services/tracker/src/tracker/s3.py
JarettForzano, line 1: I think it might be easier for you to make a class for the S3Client and make use of properties and class methods. I can tell that the way we are using these utils will fit that design pattern better since you can re-use class level attributes like `s3_client = boto3.client("s3")` and make a private attribute for `S3_BUCKET_NAME` instead of doing `bucket_name = S3_BUCKET_NAME`

contracts/claude_code/setup.sh
JarettForzano: Yes, i included it initially for testing and because i was too lazy to manually install vim for each sandbox i started

services/tracker/pyproject.toml
```diff
@@ -8,8 +8,14 @@ authors = [
...
 dependencies = [
     "boto3>=1.42.20",
-    "daytona>=0.128.1",
+    "boto3-stubs[s3]",
+    "moto[s3]",
     "fastapi[standard]>=0.127.1",
+    "ruff",
+    "pytest>=9.0.1",
+    "pytest-asyncio>=1.3.0",
+    "basedpyright",
```
JarettForzano: Lets separate dev vs production dependencies to lower the amount of packages we need to download when we deploy this service

src/agentic_harness/entry.py
JarettForzano, line 1: Is there a util you can re-use for importing the module that we use to locate the module when we upload it to s3?

cli/tracker_service.py
```diff
@@ -0,0 +1,106 @@
...
+
+import httpx
+
+from cli.config import TRACKER_URL
+
+
+class TrackerServiceError(Exception):
+    """Exception raised for tracker service errors."""
+
+    pass
```
JarettForzano, line 13: I think this is redefined

services/tracker/main.py
```diff
@@ -1,9 +1,39 @@
...
+    """Handle benchmark service communication errors"""
+    logger.error(f"Benchmark service error: {exc}", exc_info=True)
+    raise HTTPException(status_code=502, detail=str(exc)) from exc
+
+
+@app.exception_handler(S3Error)
+async def s3_error_handler(_request: Request, exc: S3Error):
+    """Handle S3 storage errors"""
+    logger.error(f"S3 error: {exc}", exc_info=True)
+    raise HTTPException(status_code=500, detail=str(exc)) from exc
```
JarettForzano: Since these are all non-generic and have the same error handling other than that log statement these should be consolidated. If you want the custom log statement you may be able to check the instance of the exception and delegate from there

services/tracker/tests/integration/test_sandbox.py
```diff
@@ -0,0 +1,179 @@
...
+    """Integration tests for sandbox operations."""
+
+    async def test_create_and_cleanup_sandbox(self, daytona: AsyncDaytona) -> None:
+        """Test that sandbox is created and cleaned up properly."""
+        sandbox_name = "test-cleanup-sandbox"
+
+        async with create_sandbox(daytona, sandbox_name, "python:3.11-slim") as sandbox:
+            assert sandbox.name == sandbox_name
+            result = await sandbox.process.exec("echo 'test'")
+            assert result.exit_code == 0
```
JarettForzano, line 102: If you are going to test cleanup after we have exited the context manager we need to check if the sandbox exists, right now we are just checking that the sandbox was created

services/tracker/tests/integration/test_sandbox.py
```diff
@@ -0,0 +1,179 @@
...
+    ) -> None:
+        """Test uploading a contract to the sandbox."""
+        # Upload contract zip to mocked S3
+        mock_s3.put_object(Bucket=S3_BUCKET_NAME, Key="contracts/test_contract.zip", Body=minimal_contract_zip)
+
+        # Upload contract to sandbox
+        await upload_contract_to_sandbox(test_sandbox, "test_contract")
+
+        # Verify contract file exists in sandbox
+        result = await test_sandbox.process.exec("test -f /bundle/contracts/test_contract/contract.py")
```
JarettForzano, line 115: What does the command `test -f ...` do?

services/tracker/tests/integration/test_sandbox.py
```diff
@@ -0,0 +1,179 @@
...
+        assert result.exit_code == 0, "Deeply nested file should exist"
+
+        result = await test_sandbox.process.exec("test -f /bundle/contracts/nested/other/file.txt")
+        assert result.exit_code == 0, "Other nested file should exist"
+
+    async def test_install_dependencies_errors_on_missing_bundle(
+        self, test_sandbox: AsyncSandbox, mock_s3: S3Client, minimal_contract_zip: bytes
+    ) -> None:
+        """Test that install_dependencies raises SandboxError when bundle is missing."""
+        from tracker.exceptions import SandboxError
```
JarettForzano: would this create a circular dependency error?

JarettForzano (changes requested):
Left some architectural feedback, after checking out all you had to do to get the contract  + submodule from point A to point B inside of the sandbox I think we should re-evaluate how we get that done. It does not seem like we are going to be able to scale this any farther as different agent architectures could break this flow.

You used the term "bundle" a few times in the code so I would like to present a solution that involves creating a executable bundle of the agent you can upload https://pyinstaller.org/en/stable/

There are a few different available and im sure we can add support for all of them pretty easily. We can chat more through implementation and direction later today during our check in. Of course feel free to give me feedback if this does not seem like a good solution

services/tracker/tests/integration/test_sandbox.py
oalmatov: utility to check if file exists

services/tracker/tests/integration/test_sandbox.py
oalmatov: removed, made just one import at the top

services/tracker/main.py
oalmatov: consolidated

services/tracker/pyproject.toml
oalmatov: yep

Makefile
oalmatov: 1. acknowledged, will do that
2. i do think that the make commands make the dev experience better because you don't have to cd into dirs to run commands
3. the unit tests make command is just so you don't have to cd into a repo to run commands.

JarettForzano: Sorry was on 4 hours of sleep and was not super clear explaining 2 and 3. I meant that the tests inside of the benchmark services are not supposed to be ran from within the tracking service. Its like if you were to run the model library tests inside of your project because you are using it as a dependency

PR #15, opened 2026-01-11
V1 Progress

JarettForzano:
### Main changes
- Feedback from previous version integrated
- Added more unit and integration tests for the tracker service
- Updated the tracker endpoints
- Added additional cli commands to have a more complete flow
- Removed model-library dependency
- Added dockerfile and docker make commands
- Updated readme
- Organized utils and unit tests
- Added types for the endpoints and doc strings

### Run setup
```bash
# Setup the cli and environment
make install

# Put these inside of the tracker/.env file
DAYTONA_API_KEY=dtn_5ebaf9926a53x
DAYTONA_API_URL=https://app.daytona.io/api
DAYTONA_TARGET=us
BENCHMARK_SERVICE_URL=http://xxx.xx.xx:8000


# from within the agentic-harness
# after this docker container is built you can run harness commands
make tracker-service

# With specific task IDs:
uv run harness start-benchmark \
  --contract <contract_path> \
  --benchmark <benchmark_name> \
  --concurrency 1 \
  --task-ids "task_1_id,task_2_id" \
  --slice "start:stop:step"

# Or run whole benchmark (not recommended for development):
uv run harness start-benchmark \
  --contract <contract_path> \
  --benchmark <benchmark_name> \
  --concurrency 1 \
  --slice "start:stop:step"


# Live updates every 60 seconds
uv run harness fetch-benchmark --benchmark-id <benchmark_id> --connect

# One-time status check
uv run harness fetch-benchmark --benchmark-id <benchmark_id>

# Download results
uv run harness retrieve-results --benchmark-id <benchmark_id> --path ./results.json
```


cli/utils.py
```diff
@@ -0,0 +1,159 @@
...
+
+def stream_benchmark_status(tracker: TrackerService, benchmark_id: UUID) -> None:
+    """
+    Stream and display live benchmark status updates.
+
+    Args:
+        tracker: TrackerService instance
+        benchmark_id: Benchmark UUID to stream
+    """
+    status_colors = {
```
JarettForzano, line 111: TODO: Can replace with constants from tracker package

services/tracker/README.md
```diff
@@ -1,13 +1,40 @@
...
+# View container logs
+make logs
+
+# Stop the container
+make stop
+
+# Clean up (remove container and image)
+make clean
+
+# Clean build and run the tracker service
```
JarettForzano: Clean, build...

services/tracker/tests/integration/test_database_integration.py
```diff
@@ -260,13 +288,19 @@ async def test_end_to_end(
             assert response == {"status": "ok"}
 
             # Create benchmark row to initiate a benchmark
-            benchmark_row = Benchmark(name=benchmark_service.name)
+            benchmark_row = Benchmark(
+                name=benchmark_service.name,
```
JarettForzano: TODO: pytest.fixture for benchmark row

services/tracker/src/tracker/database/tracker.db
oalmatov, line 1: add `tracker.db*` to gitignore?

cli/utils.py
```diff
@@ -0,0 +1,159 @@
...
+    Returns:
+        bool: True if the tracker service is healthy, False otherwise
+    """
+    response = tracker.health_check()
+    if response.status_code != 200:
+        click.echo(click.style("Tracker service failed to respond!", fg="red", bold=True))
+        click.echo(json.dumps(response.json(), indent=4, default=str))
+        return False
+
+    return True
```
oalmatov, line 26: i think we should just raise on status inside the `tracker_service.py` (something like `response.raise_on_status()` i believe). in cli main, we should just catch any `TrackerServiceError` and do this formatting.

oalmatov: unless there is a specific reason you only want to format the health check errors in this way

services/tracker/src/tracker/database/models.py
```diff
@@ -4,54 +4,110 @@
...
     FINISHED = "finished"
+    ERROR = "error"
 
 
 class TaskStatus(str, Enum):
     STARTING = "starting"
     IN_PROGRESS = "in_progress"
     EVALUATING = "evaluating"
     FINISHED = "finished"
+    ERROR = "error"
```
oalmatov, line 36: i presume finished means task completed successfully? maybe

services/tracker/src/tracker/database/models.py
```diff
@@ -4,54 +4,110 @@
...
-        """Select all evaluation results for a given benchmark"""
-        statement = (
-            select(EvaluationResult)
-            .join(Task, col(EvaluationResult.task_id) == col(Task.id))
-            .where(col(Task.benchmark_id) == self.benchmark_id)
-        )
-        return session.exec(statement)
+    @field_serializer("id", "benchmark")
+    def serialize_uuid(self, value: UUID | str) -> str:
+        return str(value)
```
oalmatov, line 48: i think this serialization is done automatically by fastapi, but i may be wrong

services/tracker/src/tracker/database/models.py
```diff
@@ -4,54 +4,110 @@
...
+    )  # TODO: Automatically set to finished when all tasks are in a finished state or error state
+
+    error_message: str | None = Field(default=None)
+    arguments: BenchmarkArguments = Field(
+        sa_column=Column(BenchmarkArgumentsType),
+    )
+
+    def fetch_final_evaluation(self, session: Session) -> FinalEvaluation | None:
+        statement = select(FinalEvaluation).where(FinalEvaluation.benchmark == self.id)
+        return session.exec(statement).first()
```
oalmatov: can we not have a foreign key for final_evaluation and just save it when we create a FinalEvaluation row.

services/tracker/src/tracker/database/models.py
```diff
@@ -4,54 +4,110 @@
...
+
+    def fetch_evaluation_results(self, session: Session) -> dict[str, dict[str, Any]]:
+        from tracker.utils import fetch_evaluation_results
+
+        return fetch_evaluation_results(self.id, session)
```
oalmatov, line 113: not sure why we have this method in both Benchmark and FinalEvaluation.

services/tracker/src/tracker/database/session.py
oalmatov, line 1: this is just for local development right?

services/tracker/src/tracker/utils.py
```diff
@@ -0,0 +1,269 @@
...
+        .where(Task.benchmark == benchmark_id)
+    )
+    results = session.exec(statement).all()
+
+    evaluation_results: dict[str, dict[str, Any]] = {}
+    for evaluation_result, task_id in results:
+        result_data = evaluation_result.result
+        evaluation_results[task_id] = result_data
+
+    return evaluation_results
```
oalmatov, line 210: same here. curious if we can define a one-to-many relationship between Benchmark and Task (i think theres something like Relationship in sqlmodel), we will also define a EvaluationResult foreign key in Task. Then either in a separate model like FinalEvaluation or inside of Benchmark we can define a property that can just do 

`evaluation_results = {task.id: task.evaluation_result for task in benchmark.tasks]`

src/agentic_harness/base/types.py
```diff
@@ -1,47 +1,43 @@
...
 from pydantic import BaseModel, model_validator
 from wonderwords import RandomWord
 
 rw = RandomWord()
 
 
-class EnvironmentKeys(BaseModel):
-    """Inherited by environments expected to return desired dependencies to inject into the sandbox environment"""
+# NOTE: Was getting some dependency issues with the model-library
+# Removing the package fixed it
```
oalmatov, line 10: Will address this in my pr

cli/utils.py
JarettForzano: I did it this so that I could have freedom to make the error messages flexible

services/tracker/src/tracker/database/models.py
JarettForzano: Its used in both situations, to fetch all evaluation results from a benchmark and all evaluation results from a final evaluation row.  I would checkout where its being called for more context

services/tracker/src/tracker/database/models.py
JarettForzano: Not really, the benchmark row is not dependent on the final evaluation result. I don't think its normal to create the relationship the way you described since we would be populating the fk value with a none value when the benchmark row is created.

services/tracker/src/tracker/database/models.py
JarettForzano: yeah, when its "finished" the task is done running

services/tracker/src/tracker/database/models.py
JarettForzano: image

could be because of SQLModel. this is just so that it knows that we want it to be a string when we serialize it so the warning goes away

services/tracker/src/tracker/database/session.py
JarettForzano: this method is called to create the database inside of the docker container. technically at some point we will substitute it for a relational database in aws (just for production) but keep that for development

services/tracker/src/tracker/utils.py
JarettForzano: what benefit does that provide over what is there currently?

services/tracker/src/tracker/database/models.py
oalmatov: in the test where we call this on FinalEvaluation, we also have access to the Benchmark row, is there a reason we have this method in both then?

services/tracker/src/tracker/database/models.py
oalmatov: i see we should just leave it then

cli/utils.py
oalmatov: but we can control the error messages from the tracker_service right? click would just style them

services/tracker/src/tracker/utils.py
```diff
@@ -0,0 +1,267 @@
...
+                        task=task_row.id, instance_id=sandbox.id, result=evaluation_result
+                    )
+                    task_session.add(evaluation_result_row)
+
+                    # Mark the task status as finished since we have finished processing the task
+                    task_row.status = TaskStatus.FINISHED
+                    task_session.add(task_row)
+                    task_session.commit()
+
+                    return evaluation_result_row, task_id
```
oalmatov: Why do we need to return a tuple and not just the row?

services/tracker/src/tracker/utils.py
```diff
@@ -0,0 +1,267 @@
...
+        )
+
+        # NOTE: Tasks with errors will still need to be included inside of the final score calculation to ensure that they are accounted for
+        evaluation_results: dict[str, dict[str, Any] | None] = {
+            task_id: evaluation_result.result if isinstance(evaluation_result, EvaluationResult) else evaluation_result
+            for evaluation_result, task_id in evaluation_result_rows
+        }
+
+        # Calculate the final score based off the tasks that were ran
+        final_score_response = await benchmark_service.request_final_score(evaluation_results=evaluation_results)
```
oalmatov, line 126: Why can't we just pass in the EvaluationResult objects into request_final_score and then model_dump them to send a request to the benchmark service? This way we don't have to have to use dict[str, dict[str, Any]]. 

Similarly in the FinalScoreResponse we can just have the EvaluationResult pydantic model

JarettForzano: Three main things
- Better interface to fetch specific task results
- Without it we cannot do things like determine which tasks passed and which ones failed (task_id is not guaranteed to be inside of the evaluation result object)
- More efficient than using a list

What do you mean by this part, `Similarly in the FinalScoreResponse we can just have the EvaluationResult pydantic model`, we do not return evaluation result objects with the final score response object.

We also don't need the whole SQLModel to send over to the service so we would not be passing in the `EvaluationResult` in any case

services/tracker/src/tracker/utils.py
JarettForzano: Its used to create the mapping between the evaluation row and the task id. going to change this to just return the mapping itself instead of doing it after.

cli/utils.py
JarettForzano: The error message from the tracker service if we could not connect would not be human readable

PR #16, opened 2026-01-14
Contract Refactor and other clean up

oalmatov:
# Contract Refactor
- Contract now defines:
    - Name of the agent
    - Paths to agent artifacts that need to be uploaded to sandbox
    - Command for install the agent
    - Command to run the agent
    - Any environment variables required to run the agent
- Bundling flow:
    - Zips up the agent artifacts and uploads them to s3
    - Tracker service downloads artifacts from s3 and uplaods them to sandbox
    - Run one install command to install dependencies
    - Run one run command to run the agent

    This is a pretty flexible approach. Artifacts can be anything, python
    packages, executables, etc, but you must define the install and run
    commands.

- Tracker Sandbox Utils:
    - Simplified sandbox upload and removed unnecessary helper functions
    - Added proper integration tests for sandbox utils

# File Structure
- Removed the weird dependency loop between client and tracker.
    - Now tracker depends on the agentic_harness package, but not vice
    versa
        - Moved all API types to the client (i.e. agentic_harness)
            - This is neat because now our client and tracker are
            guaranteed to have response/request types in sync
        - Moved logger to tracker (but can move it back)
- Moved cli into the agentic_harness package
- Renamed contracts folder to agents folder as per Langston's suggestion

# Lots of changes, mainly I would inspect `agents/claude_code`, `tracker/sandbox.py`, and `tracker/tests/integration/test_sandbox.py`


services/tracker/src/tracker/types.py
```diff
@@ -1,64 +1,41 @@
...
+    "BenchmarkArguments",
+    "BenchmarkDetails",
+    "BenchmarkStatus",
+    "FetchBenchmarkResponse",
+    "FinalEvaluationResponse",
+    "RetrieveResultsResponse",
+    "StartRunErrorResponse",
+    "StartRunRequest",
+    "StartRunResponse",
+    # SWEBench Service types
```
JarettForzano: These will be usable across all services

services/tracker/tests/utils.py
```diff
@@ -2,8 +2,16 @@
...
 
+from agentic_harness import AgentContract
 from daytona import AsyncDaytona, AsyncSandbox, CreateSandboxFromImageParams, DaytonaNotFoundError, Image, Resources
 
+TEST_CONTRACT = AgentContract(
+    name="test_contract",
+    artifacts=["setup.sh"],
+    install_cmd="bash setup.sh",
+    run_cmd="test_agent -p {{problem_statement}}",
+)
```
JarettForzano: Would this make sense to be a fixture? you can pass it into the methods without importing it and it does not need to be a const.

agents/claude_code/contract.py
```diff
@@ -0,0 +1,13 @@
...
+from agentic_harness import AgentContract
+from dotenv import load_dotenv
+
+load_dotenv()
+
+contract = AgentContract(
+    name="claude_code",
+    artifacts=["submodules/claude_code", "setup.sh"],
+    install_cmd="bash setup.sh",
+    run_cmd="claude_code -p {{problem_statement}}",
```
JarettForzano, line 11: would it be more clear if we passed the problem statement into a method instead of using string replace? would reduce user error

oalmatov: yeah let me think whats a way to do this

services/tracker/tests/utils.py
oalmatov: switching to fixture, ur right

services/tracker/src/tracker/types.py
oalmatov: comment is deleted, maybe worth decoupling benchmark service types from tracker types. but we can take care of this later

agents/claude_code/contract.py
oalmatov: will actually address this in a follow up

PR #18, opened 2026-01-18
Refactor contract/bundle flow + Implement SWEAgent + Fix claude code

oalmatov:
# Contract Refactor
- See `CONTRACTS.md` and `src/agentic_harness/contract.py` for details.

# Bundling Flow
- Zips up specified artifacts and uploads to S3.
- Tracker downloads from S3 and uploads to Daytona sandbox.
- Contract is serialized and sent with the StartRunRequest to the tracker.
- Based on contract details, the tracker runs install commands and starts the
agent.

Pretty flexible approach. Artifacts can be anything, Pthon packages,
executables, etc. As long as you define the way to install and run the agent,
you can upload anything to the sandbox.

# Run Agent changes
- Switched to using daytona PTY instaed of sessions.
- Added log streaming

# Testing
- Cleaned up and simplified test in `services/tracker/tests/integration/test_sandbox.py`
- From tracker service, run `uv run pytest tests`

# File Structure
- Tracker package now only exposes its types file.
- Removed agentic_harness dependency from tracker.
- Moved cli into agentic_harness package
- Renamed contracts folder to agents as per Langston's request.

# AgentConfig
- Added ability to pass in AgentConfig to the agent contracts.
- Currently only supports passing in model-key, but can easily be extended

# Running SWEAgent and Claude Code

Claude Code on 1 task with Haiku 4.5:
```
harness start-benchmark --agent agents/claude_code --model anthropic/claude-haiku-4-5-20251001-thinking --benchmark swebench --slice 0:1
```

SWEAgent on 1 task with Gemini 3 Flash:
```
harness start-benchmark --agent agents/sweagent --model google/gemini-3-flash-preview --benchmark swebench --slice 0:1
```


agents/claude_code/contract.py
```diff
@@ -0,0 +1,73 @@
...
+    def run_cmd(self) -> str:
+        if not self._agent_config:
+            raise ValueError("ClaudeCodeContract requires an AgentConfig")
+
+        model_name = self._agent_config.model
+        if model_name not in self._ALLOWED_MODELS:
+            raise ValueError("ClaudeCodeContract only accepts Anthropic models")
+
+        args = [
+            "-p {{problem_statement}}",
```
oalmatov: The contract is serialized when sent to the tracker service, which makes passing in a problem_statement through function arg complicated.

I can rethink this design later.

results.json
JarettForzano, line 1: Can remove, ill change the `evaluation_results` to show error's from tasks

services/tracker/src/tracker/sandbox.py
```diff
@@ -120,147 +76,79 @@ async def upload_contract_to_sandbox(sandbox: AsyncSandbox, contract_name: str)
...
-    logger.info(logs.output)
+    result = await pty_handle.wait()
 
-    if command.exit_code != 0:
-        raise SandboxError(f"Failed to run agent {contract_name} for task {task_id}")
+    if result.exit_code != 0:
+        raise SandboxError(f"Failed to run agent {contract.name}, exit code: {result.error}")
 
-    logger.info(f"Agent ran successfully for task {task_id}")
+    return ""
```
JarettForzano, line 156: The method definition seems outdated, we are only returning a blank string here

services/tracker/src/tracker/types.py
```diff
@@ -39,25 +73,26 @@ class StartRunResponse(BaseModel):
...
 
 
+class FinalEvaluationResponse(BaseModel):
+    id: UUID
+    benchmark: UUID
+    final_score: float
+    # NOTE: metadata was reserved by alchemy
+    properties: dict[str, Any] = {}
+
+
```
JarettForzano: After doing a pass on everything i got some context to what this is for, 

Is this all for this? `Tracker package now only exposes its types file.`

We are sourcing the tracker so we are going to probably be getting a lot more than just the types.py file anyway
```
[tool.uv.sources]
tracker = { path = "services/tracker", editable = true }

```
The way we did this before was good because the columns from the table were included by default. Now we are maintaining two separate versions of the same object. Also making the table inherit from an object with the same fields is a bit weird
```
class FinalEvaluation(FinalEvaluationResponse, SQLModel, table=True):
    id: UUID = Field(default_factory=uuid4, primary_key=True)
    benchmark: UUID = Field(foreign_key="benchmark.id")
    final_score: float = Field(nullable=False)
    # NOTE: metadata was reserved by alchemy
    properties: dict[str, Any] = Field(default_factory=dict, sa_column=Column(JSON, nullable=False))
```

If we are really going all in on this do not couple strategy then you should just return a json for the field inside of the response since all we are doing in the cli side of things is dumping it to a json anyway. No reason to add duplication for a typed object we won't use

`final_evaluation: dict[str, Any] | None`

JarettForzano: I would also watch the terminology since all objects being used as responses to an endpoint were suffixed "Response", here it is just a field so you should not add the suffix to it.

src/agentic_harness/cli/main.py
```diff
@@ -73,7 +82,8 @@ def start_benchmark(
     """
     click.echo("Arguments:")
     click.echo(f"  - Benchmark: {benchmark}")
-    click.echo(f"  - Contract: {contract}")
+    click.echo(f"  - Agent: {agent}")
+    click.echo(f"  - Model: {model}")
```
JarettForzano: Model is not guaranteed

services/tracker/src/tracker/database/models.py
```diff
@@ -196,3 +185,4 @@ class EvaluationResult(SQLModel, table=True):
     task: UUID = Field(foreign_key="task.id")
     instance_id: str = Field(unique=True)
     result: dict[str, Any] = Field(default_factory=dict, sa_column=Column(JSON, nullable=False))
+    agent_output: str = Field(default="")
```
JarettForzano, line 226: We need to enforce a structured output here, can you swap this to json?

src/agentic_harness/cli/tracker_service.py
```diff
@@ -1,26 +1,30 @@
...
 )
 
-from cli.config import TRACKER_URL
+from agentic_harness.cli.exceptions import TrackerServiceError
 
+load_dotenv()
 
-class TrackerServiceError(Exception):
-    """Exception raised for tracker service errors."""
+TRACKER_URL = os.environ.get("TRACKER_SERVICE_URL")
```
JarettForzano: Did you include the default url inside of the readme .env?

Makefile
```diff
@@ -84,6 +80,16 @@ test-integration: venv_check
 tracker-service:
 	cd services/tracker && make tracker-service
 
+tracker-install:
+	@echo "Installing tracker service (separate venv)..."
+	@cd services/tracker && uv venv --python $(PYTHON_VERSION)
+	@cd services/tracker && uv sync
+	@echo "✓ Tracker service installed at services/tracker/.venv"
```
JarettForzano: Should we be re-using the make command from the tracker here?

```
install:
	uv venv --python 3.12
	uv cache clean agentic-harness
	uv sync --group dev
	@echo "🎉 Done! Run 'source .venv/bin/activate' to activate the environment locally."
```

services/tracker/src/tracker/config.py
```diff
@@ -10,18 +10,37 @@
 
 load_dotenv()
 
-BENCHMARK_SERVICE_URL = os.getenv("BENCHMARK_SERVICE_URL", "http://localhost:8002")
-S3_BUCKET_NAME = os.getenv("S3_BUCKET_NAME", "agentic-harness")
-REDIS_URL = os.getenv("REDIS_URL", "redis://localhost:6379")
```
JarettForzano, line 15: What benefit does separating these out into methods provide us when the only location they should be used is in here? This also introduces more things that we need to add to the `README.md` file because instead of just providing the fallback (that we are going to require to be used anyway) we are forcing explicit behavior

services/tracker/src/tracker/types.py
```diff
@@ -1,29 +1,63 @@
...
     concurrency: int = 5
     task_ids: list[str] | None = None
     slice_str: str | None = None
 
-    @property
-    def benchmark_service(self) -> "BenchmarkService":
-        from tracker.benchmark_service import BenchmarkService
-
-        return BenchmarkService(name=self.benchmark_name, url=BENCHMARK_SERVICE_URL)
-
```
JarettForzano, line 26: Why remove this util? It was useful for instantiating the client and delegating the url we want to use. You are also now forcing a combination of helper methods that essentially do the same thing but require more steps

```
url = get_benchmark_service_url()
benchmark_service = BenchmarkService(name=start_run_request.benchmark_name, url=url)
```

services/tracker/tests/integration/test_process_benchmark.py
```diff
@@ -61,22 +61,28 @@ async def _test_run_agent(self, original_method: Any, *args: Any, **kwargs: Any)
 
         import tracker.utils
 
-        task_id: str = args[2]
+        # The sandbox name is the task_id
+        sandbox = args[0]
+        task_id: str = sandbox.name
```
JarettForzano: Why swap it to use the sandbox name? This will just break when we add unique names to the sandbox because we are using the wrong variable to source the task from the db

src/agentic_harness/cli/main.py
```diff
@@ -86,22 +96,30 @@ def start_benchmark(
         formatted_task_ids = task_ids.split(",")
         click.echo(f"Discovered {len(formatted_task_ids)} task IDs")
 
-    click.echo("Validating contract...", nl=False)
-    bundler.validate_contract(contract)
-
     try:
+        contract_path = agent / "contract.py"
+        agent_config = AgentConfig(model=model) if model else None
```
JarettForzano: I originally thought that the reason we made the agent config was to have more options but here it seems pretty coupled with the model field. 

If we are going to have a config usually you make it required and just populate it with empty values if its not in use. That way the agent config is not dependent on if the model field exists inside. It will be easier for you to use

src/agentic_harness/schemas.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class AgentConfig(BaseModel):
+    """Configuration for the agent."""
+
+    model: str
+    """Model key (e.g., openai/gpt-4o)"""
+
+
+class BenchmarkConfig(BaseModel):
```
JarettForzano: Could not find any references to this

services/tracker/src/tracker/sandbox.py
oalmatov: I'm currently returning blank string because we haven't decided how we'll save the output

JarettForzano (commented):
Changing the config to just have helper methods caused the unit tests to break since we are enforcing environment variables to be soured that are not required. I also don't see how its better than making them constants with defaults.

Removing the BenchmarkService instantiate helper method from the Benchmark class also introduced some code smells since we are duplicating the logic required to instantiate the service.

In the future we could have delegated the url inside of the class depending on what the benchmark name was. I left some comments so maybe after explaining ill understand

src/agentic_harness/schemas.py
oalmatov: removed

src/agentic_harness/cli/main.py
oalmatov: added a check

src/agentic_harness/cli/main.py
oalmatov: noted, added defaults (just None for now) and constructing config by unpacking kwargs

services/tracker/src/tracker/types.py
oalmatov: siwtched back to getting from envs with defaults

services/tracker/src/tracker/types.py
oalmatov: yeah removed the `include-only` part from pyproject and just kept a unified FinalEvaluation type

src/agentic_harness/cli/tracker_service.py
oalmatov: switched back to jsut usng defaults

services/tracker/src/tracker/config.py
oalmatov: added back the fallback

services/tracker/tests/integration/test_process_benchmark.py
oalmatov: added back task_id as an arg to run_agent

JarettForzano (approved):

Makefile
oalmatov: cleaned up makefiles

services/tracker/src/tracker/database/models.py
oalmatov: my todo for v2 is to have a consistent dir to store logs and agent outputs in the sandbox. i want to download this dir and save it to s3 and store the s3 key to the database. for now lets leave it as is.

PR #19, opened 2026-01-19
Added table view for displaying models

JarettForzano:
### Main changes
- Added endpoint `@app.get("/fetch-benchmarks")`
- Added table view to allow users to browse benchmarks ran
- Updated documentation
- Added fastapi unit tests for new endpoint

### Testing info

If you want to test this feature out, use this
1. Add at the root of agentic harness  https://pastebin.com/K9x9kSnC (seed.py)
2. Start tracking service `make tracker-service`
3. Copy file inside `docker cp services/tracker/seed_data.py tracker-service-container:/app/seed.py`
4. Run the script to populate the tables with random info `docker exec tracker-service-container uv run python seed.py`

NOTE: Check the readme for complete list of filter options

```bash
uv run harness fetch-benchmarks --contract-name <contract_name> --benchmark-name <benchmark_name> 
```


src/agentic_harness/cli/main.py
```diff
@@ -289,5 +291,56 @@ def resume_run(benchmark_id: UUID):
...
+    required=False,
+    default=Order.DESC.value,
+    help="Order by the benchmarks to fetch (e.g., desc, asc)",
+)
+def fetch_benchmarks(
+    contract_name: str | None,
+    benchmark_name: str | None,
+    status: str | None,
+    order_by: str = "desc",
+):
```
oalmatov, line 326: do u think we should rename this to something like list benchmarks?

in benchmarks-be i remember langston complaining about having two scripts run-benchmark and run-benchmarks which was confusing

services/tracker/tests/unit/test_fastapi_server.py
```diff
@@ -405,3 +405,116 @@ def get_test_session():
...
+        fetch_benchmarks_request.status = BenchmarkStatus.IN_PROGRESS
+
+        # When we fetch with benchmarks found, we return a 200 OK
+        response = client.get(
+            "/fetch-benchmarks", params=fetch_benchmarks_request.model_dump(exclude_none=True, mode="json")
+        )
+        assert response.status_code == 200
+        response_json = response.json()
+        assert response_json.get("total_count") == 5
+        assert len(response_json.get("benchmarks")) == 5
```
oalmatov: why are we searching for 4 but asserting 5?

services/tracker/src/tracker/utils.py
```diff
@@ -585,3 +587,43 @@ async def resume_benchmark(
...
+
+    Returns:
+        tuple[Sequence[Benchmark], int]
+        Sequence of benchmark rows and total count of benchmark rows
+
+    """
+    query = select(Benchmark)
+
+    if request.contract_name:
+        query = query.where(func.json_extract(Benchmark.arguments, "$.contract.name") == request.contract_name)
```
oalmatov, line 607: @JarettForzano flagging this is one line i had to change. since we are now saving AgentContractRequest to the db

services/tracker/tests/unit/test_fastapi_server.py
JarettForzano: before creating the 4 i made a benchmark row

```
 # Add benchmark row to the database to fetch
database_session.add(example_benchmark_object)
database_session.commit()
```

the comment is wrong- good catch ill fix it

src/agentic_harness/cli/main.py
JarettForzano: if i remember correctly i had the same complaint, it was because the scripts were in two different files and the script to run a single benchmark was not necessary. here its different since if you fetch a single benchmark the view is different then when you want to fetch multiple benchmarks. 

might sound funny but i don't really like terms that don't roll of the tongue well, its why i avoid "list" in preference to "fetch"

running a single benchmark falls under the same logic as running multiple benchmarks which is why the script to run a single benchmark was a problem since its just redundant.


TLDR: The word "list" does not sound as good as "fetch" 😆

oalmatov: makes sense, lets keep it

oalmatov (approved):

PR #20, opened 2026-01-19
Give sandboxes unique names

JarettForzano:
### Main changes
- Added a suffix to the end of a sandbox name. 

### Bug notes
Daytona has a unique restriction on sandbox names. We did not have a suffix so if we tried to run the same benchmark on different models it would error out.

### Solution
Add a suffix to the sandbox name that can be shared across all tasks inside of a benchmark

`uv run pytest tests/integration/test_process_benchmark.py::TestProcessBenchmark::test_can_run_same_task`


services/tracker/src/tracker/sandbox.py
```diff
@@ -80,13 +80,18 @@ def get_submodules_dir(contract_name: str) -> PurePosixPath:
...
+) -> AsyncGenerator[AsyncSandbox, Any]:
     """
     Create a sandbox with the given name and image.
     Automatically cleans up the sandbox when the context manager exits.
     """
     logger.info(f"Creating sandbox {sandbox_name} with image {image}")
 
+    if hash_suffix:
+        sandbox_name = f"{sandbox_name}-{hash_suffix}"
+
```
oalmatov, line 50: do you think we should relate it to the benchmark run id?

JarettForzano: I used a hash since i wanted to keep the suffix short and use letters. The hash is made using the benchmark id as a part of it

oalmatov: yeah i thought it would be useful for searching for sandboxes related to the benchmark run through daytona, but we could do that on the cli side too

JarettForzano: yeah, we don't have a way to see sandbox info from the util now but in the future we should do it. I could imagine you can scroll through the benchmarks and after selecting one see all the tasks are running and be able to enter it and see the logs from the cli

PR #29, opened 2026-01-22
Added cwd to the tracker

JarettForzano:
### Main changes
- Added cwd from the benchmark service to the tracker (serves as the command work directory we execute the agent in)
- Fixed the bug with the setup-task url (was failing with the slash)
- Handle errors better in unit tests

### Usage
Add this to the `.env` 

`BENCHMARK_SERVICE_URL=https://benchmark-service.vals.ai`

### Tests
All tests pass

### DIFF
--- a/services/tracker/src/tracker/benchmark_service.py
+++ b/services/tracker/src/tracker/benchmark_service.py
@@ -144,7 +144,7 @@ class BenchmarkService:
 
         async with httpx.AsyncClient(timeout=None, follow_redirects=True) as client:
             response = await client.post(
-                f"{self._url}/setup-task/",
+                f"{self._url}/setup-task",
                 json={"task_id": task_id, "instance_id": instance_id},
                 headers={
                     "Content-Type": "application/json",
--- a/services/tracker/src/tracker/utils.py
+++ b/services/tracker/src/tracker/utils.py
@@ -213,7 +213,9 @@ async def process_task(
 
                 # Run the agent inside of the sandbox
                 # NOTE: Currently only testing when agent does not need a response, in the future run agent will return a json to evaluate it needed
-                await run_agent(sandbox, start_run_request.contract, task_data.problem_statement, task_id)
+                await run_agent(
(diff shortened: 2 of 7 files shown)


oalmatov (approved):

services/tracker/src/tracker/benchmark_service.py
```diff
@@ -144,7 +144,7 @@ async def request_setup_task(self, task_id: str, instance_id: str) -> SetupTaskR
 
         async with httpx.AsyncClient(timeout=None, follow_redirects=True) as client:
             response = await client.post(
-                f"{self._url}/setup-task/",
+                f"{self._url}/setup-task",
```
oalmatov, line 147: what's this for?

services/tracker/src/tracker/utils.py
```diff
@@ -213,7 +213,9 @@ async def process_task(
 
                 # Run the agent inside of the sandbox
                 # NOTE: Currently only testing when agent does not need a response, in the future run agent will return a json to evaluate it needed
-                await run_agent(sandbox, start_run_request.contract, task_data.problem_statement, task_id)
+                await run_agent(
+                    sandbox, start_run_request.contract, task_data.problem_statement, task_id, task_data.cwd
+                )
```
oalmatov, line 218: should we just pass in the full task_data object instead of unpacking the problem_statement, task_id, and cwd?

services/tracker/src/tracker/benchmark_service.py
JarettForzano: The benchmark service defined the endpoint without the slash and it was here for some reason. It caused the redirect to fail and the endpoint to not be reachable. It worked before so it could have been a side affect to some of the more recent changes that was made with https but not really important. This fixed the tests

services/tracker/src/tracker/utils.py
JarettForzano: i don't think it matters much, we are actually only using the problem statement and the cwd from the object. the task id is not from the request. Its just more clear when the params are explicit

PR #31, opened 2026-01-22
Upgraded development environment

JarettForzano:
### Main changes
- Separated services to use docker-compose
- Change make file to use docker compose commands
- Removed depreciated create database command from readme and commands (now created using alembic)
- Fixed bug where dependencies were being downloaded twice due to supervisord
- Changed install command to also init the database

### Previously
- Database was reset each time we restarted the container
- Had more commands in readme than necessary (one command does it all)
- Had to run session script manually to create database (now its automated through make)

### DIFF
--- a/services/tracker/supervisord.conf
+++ b/services/tracker/supervisord.conf
@@ -3,7 +3,7 @@ nodaemon=true
 user=root
 
 [program:api]
-command=uv run uvicorn main:app --host 0.0.0.0 --port 8000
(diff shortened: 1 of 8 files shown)


services/tracker/supervisord.conf
JarettForzano, line 1: Before uv was re-installing all the packages we already had installed inside of the container. Using no sync lets us re-use the packages we installed

oalmatov (approved):

PR #33, opened 2026-01-26
Force stop feature

JarettForzano:
### Main changes
- Refactored the way we were handling task alias and contained it to just the task context (easier to make and use)
- Added a force flag to the `stop_run` method that initiates force closure of sandboxes on the users request
- Added an alias property to Task table that allows us to create an alias for a task that we can use to make sandbox names unique (replaces older way)
- Added labels to sandboxes, this holds information we can use to create a generator and fetch the sandboxes up front
- Refactored the `process_task` method to handle force sandbox exit
- Reverted regression with `Image.Base(image_name)` syntax, should always be just the image string

### Unit tests
- Updated syntax for new --force flag
- Removed raise test for when we stop a benchmark and all tasks are running (since now we can force stop past this point)
- Added a new test for testing force stop functionality `tests/integration/test_force_stop.py`

### Test run instructions
`uv run pytest tests/integration/test_force_stop.py -s`

### New --force flag example
`uv run harness stop-run --benchmark-id <benchmark_id> --force`

### Additional Comments
- I would still consider this beta but the functionality works
- Noticing some fluctuations with the live watch feature. Ill fix that in another pr

### DIFF
--- a/services/tracker/src/tracker/benchmark_service.py
+++ b/services/tracker/src/tracker/benchmark_service.py
@@ -23,6 +23,7 @@ class BenchmarkService:
     _name: str
     _url: str
     _environment_keys: dict[str, str]
+    _daytona_client: AsyncDaytona | None = None
 
     def __init__(self, name: str, url: str):
         self._name = name
@@ -39,12 +40,18 @@ class BenchmarkService:
 
     @property
     def daytona_client(self) -> AsyncDaytona:
+        if self._daytona_client:
+            return self._daytona_client
+
         config = DaytonaConfig(
             api_key=self.environment_keys["DAYTONA_API_KEY"],
             api_url=self.environment_keys["DAYTONA_API_URL"],
             target=self.environment_keys["DAYTONA_TARGET"],
         )
-        return AsyncDaytona(config=config)
+
+        self._daytona_client = AsyncDaytona(config=config)
+
+        return self._daytona_client
 
     @staticmethod
     def daytona_keys() -> dict[str, str]:
--- a/services/tracker/src/tracker/sandbox.py
+++ b/services/tracker/src/tracker/sandbox.py
@@ -159,34 +159,31 @@ async def run_agent(
 
     session_id = f"{contract.name}-{task_id.replace(' ', '_')}"
 
-    try:
-        await sandbox.process.create_session(session_id)
+    await sandbox.process.create_session(session_id)
 
-        session_exec_resp = await sandbox.process.execute_session_command(
-            session_id, SessionExecuteRequest(command=f"cd {cwd} && {run_cmd}", run_async=True)
+    session_exec_resp = await sandbox.process.execute_session_command(
+        session_id, SessionExecuteRequest(command=f"cd {cwd} && {run_cmd}", run_async=True)
+    )
+
+    cmd_id = session_exec_resp.cmd_id
+
+    if not cmd_id:
+        raise SandboxError(f"Failed to execute command {run_cmd} in session {session_id}")
+
+    log_task = asyncio.create_task(
+        sandbox.process.get_session_command_logs_async(
+            session_id=session_id,
+            command_id=cmd_id,
+            on_stdout=on_data,
+            on_stderr=on_data,
         )
+    )
 
-        cmd_id = session_exec_resp.cmd_id
+    await log_task
 
-        if not cmd_id:
-            raise SandboxError(f"Failed to execute command {run_cmd} in session {session_id}")
+    cmd = await sandbox.process.get_session_command(session_id, cmd_id)
 
-        log_task = asyncio.create_task(
-            sandbox.process.get_session_command_logs_async(
-                session_id=session_id,
-                command_id=cmd_id,
-                on_stdout=on_data,
-                on_stderr=on_data,
-            )
-        )
+    if cmd.exit_code != 0:
+        raise SandboxError(f"Failed to run agent {contract.name}, exit code: {cmd.exit_code}")
 
-        await log_task
-
-        cmd = await sandbox.process.get_session_command(session_id, cmd_id)
-
-        if cmd.exit_code != 0:
-            raise SandboxError(f"Failed to run agent {contract.name}, exit code: {cmd.exit_code}")
-
-        return ""
-    finally:
-        await sandbox.process.delete_session(session_id)
+    return ""
--- a/services/tracker/src/tracker/sandbox.py
+++ b/services/tracker/src/tracker/sandbox.py
@@ -13,7 +13,6 @@ from daytona import (
     AsyncSandbox,
     CreateSandboxFromImageParams,
     FileUpload,
-    Image,
     Resources,
     SessionExecuteRequest,
 )
@@ -41,22 +40,20 @@ async def create_sandbox(
     daytona: AsyncDaytona,
     sandbox_name: str,
     image: str,
-    hash_suffix: str | None = None,
+    labels: dict[str, str] | None = None,
     env_vars: dict[str, str] | None = None,
 ) -> AsyncGenerator[AsyncSandbox, Any]:
     """
-    Create a sandbox with the given name and image.
+    Create a sandbox with the given name, image, and labels.
     Automatically cleans up the sandbox when the context manager exits.
     """
     logger.info(f"Creating sandbox {sandbox_name} with image {image}")
 
-    if hash_suffix:
-        sandbox_name = f"{sandbox_name}-{hash_suffix}"
-
     sandbox = await daytona.create(
         CreateSandboxFromImageParams(
             name=sandbox_name,
-            image=Image.base(image),
+            labels=labels,
+            image=image,
             network_block_all=False,
             resources=Resources(
                 cpu=4,
--- a/services/tracker/src/tracker/sandbox.py
+++ b/services/tracker/src/tracker/sandbox.py
@@ -14,6 +14,7 @@ from daytona import (
     CreateSandboxFromImageParams,
     FileUpload,
     Resources,
+    SandboxState,
     SessionExecuteRequest,
 )
 
@@ -68,9 +69,9 @@ async def create_sandbox(
     try:
         yield sandbox
     finally:
-        logger.info(f"Deleting sandbox {sandbox.name}")
-        await daytona.delete(sandbox)
-        await daytona.close()
+        await sandbox.refresh_data()
+        if sandbox.state not in [SandboxState.DESTROYING, SandboxState.DESTROYED]:
+            await daytona.delete(sandbox)
 
 
 async def upload_agent_artifacts(sandbox: AsyncSandbox, contract: AgentContractRequest) -> None:
--- a/services/tracker/src/tracker/sandbox.py
+++ b/services/tracker/src/tracker/sandbox.py
@@ -159,31 +159,38 @@ async def run_agent(
 
     session_id = f"{contract.name}-{task_id.replace(' ', '_')}"
 
-    await sandbox.process.create_session(session_id)
+    try:
+        await sandbox.process.create_session(session_id)
 
-    session_exec_resp = await sandbox.process.execute_session_command(
-        session_id, SessionExecuteRequest(command=f"cd {cwd} && {run_cmd}", run_async=True)
-    )
-
-    cmd_id = session_exec_resp.cmd_id
-
-    if not cmd_id:
-        raise SandboxError(f"Failed to execute command {run_cmd} in session {session_id}")
-
-    log_task = asyncio.create_task(
-        sandbox.process.get_session_command_logs_async(
-            session_id=session_id,
-            command_id=cmd_id,
-            on_stdout=on_data,
-            on_stderr=on_data,
+        session_exec_resp = await sandbox.process.execute_session_command(
+            session_id, SessionExecuteRequest(command=f"cd {cwd} && {run_cmd}", run_async=True)
         )
-    )
 
-    await log_task
+        cmd_id = session_exec_resp.cmd_id
 
-    cmd = await sandbox.process.get_session_command(session_id, cmd_id)
+        if not cmd_id:
+            raise SandboxError(f"Failed to execute command {run_cmd} in session {session_id}")
 
-    if cmd.exit_code != 0:
-        raise SandboxError(f"Failed to run agent {contract.name}, exit code: {cmd.exit_code}")
+        log_task = asyncio.create_task(
+            sandbox.process.get_session_command_logs_async(
+                session_id=session_id,
+                command_id=cmd_id,
+                on_stdout=on_data,
+                on_stderr=on_data,
+            )
+        )
 
-    return ""
+        await log_task
+
+        cmd = await sandbox.process.get_session_command(session_id, cmd_id)
+
+        if cmd.exit_code != 0:
+            raise SandboxError(f"Failed to run agent {contract.name}, exit code: {cmd.exit_code}")
+
+        return ""
+    finally:
+        try:
+            await sandbox.process.delete_session(session_id)
+        except Exception:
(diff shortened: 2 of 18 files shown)


services/tracker/src/tracker/utils.py
JarettForzano, line 1: This utils file is getting chunky, in another pr ill separate it out into a folder with different util files

services/tracker/src/tracker/benchmark_service.py
```diff
@@ -39,12 +40,18 @@ def environment_keys(self) -> dict[str, str]:
 
     @property
     def daytona_client(self) -> AsyncDaytona:
+        if self._daytona_client:
```
JarettForzano, line 43: We share a single client across the whole benchmark runner now, easier to close the client at the end since it only needs to be done once.

services/tracker/src/tracker/sandbox.py
```diff
@@ -36,27 +37,38 @@ def get_contract_path(contract_name: str) -> PurePosixPath:
     return bundle_path / contract_name
 
 
+async def delete_sandbox(sandbox: AsyncSandbox, daytona: AsyncDaytona) -> None:
```
JarettForzano, line 40: Sandbox no longer is guaranteed to exist when we exit the context manager. For the record when we stop a sandbox we do not exit the context manager, it is a different instance.

services/tracker/src/tracker/sandbox.py
```diff
@@ -153,7 +163,7 @@ def on_data(data: str) -> None:
         await sandbox.process.create_session(session_id)
 
         session_exec_resp = await sandbox.process.execute_session_command(
-            session_id, SessionExecuteRequest(command=f"cd {cwd} && {run_cmd}", runAsync=True)
+            session_id, SessionExecuteRequest(command=f"cd {cwd} && {run_cmd}", run_async=True)
```
JarettForzano, line 167: Syntax change from the latest version of daytona

services/tracker/src/tracker/utils.py
```diff
@@ -169,16 +168,47 @@ def handle_early_exit(task_row: Task, task_session: Session) -> None:
...
+        A dictionary with the task id as the key and the evaluation result as the value, or None if the task was stopped early (not forced)
+
+    Raises:
+        TrackerServiceError: If the result is not the expected type
+        Exception: Unexpected error occured that was not handled
+    """
     with Session(bind=engine, expire_on_commit=False) as task_session:
         benchmark_row = fetch_benchmark_row(benchmark_id, task_session)
 
+        # Merge the task row to get the latest state from the database
```
JarettForzano, line 187: Since we create the task outside of this session we need to merge the task row passed in. Now we don't need to do this anywhere else unlike before

services/tracker/src/tracker/utils.py
```diff
@@ -187,63 +217,74 @@ async def process_task(
...
+                    task_session.commit()
+
+                    return {task_id: evaluation_result_row.result}
+                except Exception as e:
+                    # Validate that the sandbox has not been destroyed
+                    task_session.refresh(task_row)
+                    if task_row.status == TaskStatus.STOPPED and await check_force_stop(sandbox):
+                        return {task_id: None}
+
+                    raise e from e
```
JarettForzano, line 259: This is the only part added here, if the task was stopped and the sandbox no longer exists we know that the exception was caused by the client being unable to connect to the sandbox.

services/tracker/src/tracker/utils.py
```diff
@@ -537,7 +580,7 @@ async def stream_benchmark_results(benchmark_id: UUID, session: Session) -> Asyn
         yield DISCONNECT
 
 
-async def stop_benchmark(benchmark_row: Benchmark, session: Session) -> None:
+async def stop_benchmark(benchmark_row: Benchmark, session: Session, force: bool) -> None:
```
JarettForzano, line 564: This section works the same as before besides this time if we are force stopping a benchmark we want to update the status even if tasks are being processed. This is because they will be killed shortly after

services/tracker/tests/unit/test_benchmark_utils.py
```diff
@@ -121,26 +121,9 @@ def get_test_session():
         database_session.commit()
 
         # Fail to stop the run
-        response: Response = client.post(f"/stop-run/{benchmark_row.id}")
+        response: Response = client.post(f"/stop-run/{benchmark_row.id}?force=false")
         assert response.status_code == 400
 
-        # Cannot stop a run with no tasks yet to start
```
JarettForzano, line 127: We no long raise here

services/tracker/src/tracker/utils.py
```diff
@@ -572,12 +610,109 @@ async def stop_benchmark(benchmark_row: Benchmark, session: Session) -> None:
...
+            if sandbox.state in [SandboxState.DESTROYING, SandboxState.DESTROYED]:
+                continue
+
+            yield sandbox
+
+        # NOTE: Since we deleted the first 10 the first page will be populated with the next 10 sandboxes
+        paginated_sandboxes = await fetch_sandboxes(benchmark_row, daytona_client, int(paginated_sandboxes.page))
+
+
+async def force_stop_sandboxes(benchmark_row: Benchmark, session: Session) -> None:
```
JarettForzano, line 645: This process is pretty simple, we go through the sandboxes that exist and remove the ones that are connected to the benchmark we are stopping. We filter using the tag we now add to the sandbox on creation. 

We also organize the errors and raise after we try to close all of the sandboxes.

services/tracker/src/tracker/utils.py
```diff
@@ -572,12 +610,109 @@ async def stop_benchmark(benchmark_row: Benchmark, session: Session) -> None:
...
+        return f"{str(e)}: {traceback.format_exc()}"
+
+
+async def fetch_sandboxes(benchmark_row: Benchmark, daytona_client: AsyncDaytona, page: int) -> AsyncPaginatedSandboxes:
+    return await daytona_client.list(
+        labels={"Benchmark": benchmark_row.name, "Id": str(benchmark_row.id)}, limit=10, page=page
+    )
+
+
+async def sandbox_generator(
```
JarettForzano, line 618: Comment does a good job explaining here

services/tracker/src/tracker/utils.py
```diff
@@ -572,12 +610,109 @@ async def stop_benchmark(benchmark_row: Benchmark, session: Session) -> None:
             .values(status=TaskStatus.STOPPED)
         )
         session.commit()
-    except TrackerServiceError:
-        raise
     except Exception as e:
         raise TrackerServiceError(f"Unexpected error stopping benchmark {benchmark_row.id}: {str(e)}") from e
 
 
+async def stop_sandbox(
```
JarettForzano, line 592: After we delete the task successfully we change the task to stopped. This prevents us from mismatching the status with the state of the sandbox.

services/tracker/src/tracker/benchmark_service.py
oalmatov: this is good

services/tracker/src/tracker/sandbox.py
oalmatov: nice

services/tracker/src/tracker/sandbox.py
```diff
@@ -179,4 +189,8 @@ def on_data(data: str) -> None:
 
         return ""
     finally:
-        await sandbox.process.delete_session(session_id)
+        try:
+            await sandbox.process.delete_session(session_id)
+        except Exception:
+            # NOTE: If we kill the sandbox this sometimes errors
+            pass
```
oalmatov, line 198: lets add a log

services/tracker/src/tracker/sandbox.py
```diff
@@ -36,27 +37,38 @@ def get_contract_path(contract_name: str) -> PurePosixPath:
...
 
+async def delete_sandbox(sandbox: AsyncSandbox, daytona: AsyncDaytona) -> None:
+    """Delete sandbox if it is not already destroyed or being destroyed"""
+    try:
+        await sandbox.refresh_data()
+        if sandbox.state not in [SandboxState.DESTROYING, SandboxState.DESTROYED]:
+            await daytona.delete(sandbox)
+    except DaytonaNotFoundError:
+        # If we error here that means the sandbox has just been deleted before we could refresh the state
+        pass
```
oalmatov, line 49: lets add a log

services/tracker/src/tracker/sandbox.py
oalmatov: i want to be able to circle back to this and see if this is something that actually happens once we start using the harness, and if it doesn't then clean this up.

services/tracker/src/tracker/utils.py
```diff
@@ -530,14 +551,14 @@ async def stream_benchmark_results(benchmark_id: UUID, session: Session) -> Asyn
                     yield EVENT_COMPLETE
                     break
 
-            await asyncio.sleep(60)
+            await asyncio.sleep(5)
```
oalmatov: magic number, add it to the constants at the top of the function?

services/tracker/src/tracker/utils.py
```diff
@@ -554,16 +575,19 @@ async def stop_benchmark(benchmark_row: Benchmark, session: Session) -> None:
             .where(col(Task.status) == TaskStatus.STARTING)
         ).one()
 
-        if not tasks:
-            raise TrackerServiceError(
-                f"All tasks for benchmark {benchmark_row.id} have been started. Must wait for the tasks to finish running."
-            )
+        # If we are forcing we want to update the benchmark status to stopping anyway
+        if not tasks and not force:
+            return None
```
oalmatov: why are we not raising here?

oalmatov: i think i understand:

if benchmark has no tasks that are starting and we are not forcing a stop, return early.

if there are tasks or we are forcing them to stop. We are marking the benchmark status to STOPPING.

after we've marked it as STOPPING, if there are no tasks return early, otherwise switch them to STOPPED.

oalmatov: Can't this function be simplified to just this

```
  # Always stop STARTING tasks
  result = session.exec(
      update(Task)
      .where(col(Task.benchmark) == benchmark_row.id)
      .where(col(Task.status) == TaskStatus.STARTING)
      .values(status=TaskStatus.STOPPED)
  )
  session.commit()

  # Set benchmark to stopping if we stopped tasks OR if forcing
  if result.rowcount > 0 or force:
      benchmark_row.status = BenchmarkStatus.STOPPING
      session.add(benchmark_row)
      session.commit()
```

oalmatov: this allows us to just have 1 query. the update is atomic. and simpler with less lines of code.

services/tracker/src/tracker/utils.py
```diff
@@ -572,12 +596,109 @@ async def stop_benchmark(benchmark_row: Benchmark, session: Session) -> None:
...
+    task_rows: Sequence[Task] = session.exec(
+        select(Task)
+        .where(col(Task.benchmark) == benchmark_row.id)
+        .where(col(Task.status).in_([TaskStatus.IN_PROGRESS, TaskStatus.EVALUATING]))
+    ).all()
+
+    # Create a mapping upfront that we can use to update the task rows after stopping the sandboxes
+    task_metadata: dict[str, str] = {task.alias: str(task.id) for task in task_rows}
+
+    # Iterate through each running sandbox and stop it, collecting erorr messages
```
oalmatov: typo

services/tracker/src/tracker/utils.py
```diff
@@ -572,12 +596,109 @@ async def stop_benchmark(benchmark_row: Benchmark, session: Session) -> None:
...
+    results: dict[str, str | None] = {}
+    async for sandbox in sandbox_generator(benchmark_row, daytona_client):
+        # If task id does not exist, the task is finished and we can just close the sandbox
+        # Edge case where the sandbox is hanging
+        task_id = task_metadata.get(sandbox.name)
+        if not task_id:
+            logger.info(f"Discovered sandbox not being tracked with the name {sandbox.name}. Closing sandbox.")
+
+        result = await stop_sandbox(sandbox, daytona_client, session, UUID(task_id) if task_id else None)
```
oalmatov: just store UUIDs directly in the task_metadata mapping

services/tracker/src/tracker/utils.py
```diff
@@ -572,12 +596,109 @@ async def stop_benchmark(benchmark_row: Benchmark, session: Session) -> None:
...
+
+    # Create a mapping between all task primary key id and their task_id (that are pending or evaluating)
+    task_rows: Sequence[Task] = session.exec(
+        select(Task)
+        .where(col(Task.benchmark) == benchmark_row.id)
+        .where(col(Task.status).in_([TaskStatus.IN_PROGRESS, TaskStatus.EVALUATING]))
+    ).all()
+
+    # Create a mapping upfront that we can use to update the task rows after stopping the sandboxes
+    task_metadata: dict[str, str] = {task.alias: str(task.id) for task in task_rows}
```
oalmatov: nit: but we can do smth like this:

```
  # Select only alias and id columns
  task_mapping = session.exec(
      select(Task.alias, Task.id)
      .where(col(Task.benchmark) == benchmark_row.id)
      .where(col(Task.status).in_([TaskStatus.IN_PROGRESS, TaskStatus.EVALUATING]))
  ).all()

  # Convert directly to dict
  task_metadata: dict[str, UUID] = dict(task_mapping)
```

feel free to leave as is tho, i'm just being pedantic.

services/tracker/src/tracker/utils.py
```diff
@@ -572,12 +596,109 @@ async def stop_benchmark(benchmark_row: Benchmark, session: Session) -> None:
...
+    for sandbox_name, error_message in results.items():
+        if error_message:
+            error_mapping[sandbox_name] = error_message
+
+    if error_mapping:
+        error_message = "\n".join(
+            [f"{task_alias}: {error_message}" for task_alias, error_message in error_mapping.items()]
+        )
+
+        raise TrackerServiceError(f"Unexpected errors stopping sandboxes:\n{error_message}")
```
oalmatov: more pythonic way:

```
    error_message = "\n".join(
        f"{task_alias}: {error_message}" for task_alias, error_message in results.items() if error_message
    )

    if error_message:
        raise TrackerServiceError(f"Unexpected errors stopping sandboxes:\n{error_message}")
```

src/agentic_harness/cli/utils.py
```diff
@@ -51,7 +51,7 @@ def create_progress_bar(finished_tasks: int, total_tasks: int, bar_width: int =
         progress_pct = (finished_tasks / total_tasks * 100) if total_tasks > 0 else 0
         filled_width = int(bar_width * progress_pct / 100)
         bar = "█" * filled_width + "░" * (bar_width - filled_width)
-        return bar, progress_pct
+        return str(bar), progress_pct
```
oalmatov: i don't think this is necessary

JarettForzano: yeah its not, basedpyright showed me a syntax error so i changed it. if it does not show up for you i can revert

services/tracker/src/tracker/utils.py
JarettForzano: yeah that makes sense

services/tracker/src/tracker/utils.py
JarettForzano: alias is a computed field which is why i fetched the object before creating the mapping. SQLModel does not have access to it

JarettForzano:

- #33 ◀
    - #34
        - #35
            - #36
                - #37
                    - #38
                        - #39

oalmatov (approved):

PR #34, opened 2026-01-28
Resume retry enhancement

JarettForzano:
### Main changes
- Added a --retry flag to the resume command that retries all tasks with the status error (ontop of resuming the stopped tasks)
- Added unit test to ensure the status change took affect
- Updated tests and references with new flag (plus docs)

### Additional Notes
This change was pretty easy to make since the `process_benchmark` method is flexible in the way that it runs all `task_ids` that are passed in. So as long as we are resetting the tasks we can retry any tasks we want.

### Next steps
- Adding a --force flag that restarts a task in a finished state, this will be for if we ever want to re-run a set of tasks and not have to redo the whole benchmark

### DIFF
--- a/services/tracker/src/tracker/utils.py
+++ b/services/tracker/src/tracker/utils.py
@@ -710,10 +710,13 @@ async def resume_benchmark(
                 f"No tasks for benchmark {benchmark_row.id} can be resumed because all tasks are finished"
             )
 
+        # id is task row primary key, task_id is the task id
+        task_mapping: dict[str, str] = {id: task_id for id, task_id in task_ids}
+
         # Verify the task ids are still valid before priming to resume
         # Raises if any task ids are invalid
         verify_response = await benchmark_service.request_verify_task_ids(
-            task_ids=[task_id for _, task_id in task_ids], slice_str=None
+            task_ids=list(task_mapping.values()), slice_str=None
         )
 
         # Set the benchmark status to in progress to flag resuming the benchmark
@@ -734,7 +737,7 @@ async def resume_benchmark(
         )
 
         # Delete all evaluation results for the tasks (unlikely they exist)
-        session.exec(delete(EvaluationResult).where(col(EvaluationResult.task).in_([id for id, _ in task_ids])))
+        session.exec(delete(EvaluationResult).where(col(EvaluationResult.task).in_(list(task_mapping.keys()))))
 
         session.commit()
 
--- a/services/tracker/src/tracker/utils.py
+++ b/services/tracker/src/tracker/utils.py
@@ -703,7 +703,7 @@ async def resume_benchmark(
         filter_query = (col(Task.benchmark) == benchmark_row.id, col(Task.status).in_(retry_statuses))
 
         # Check if there are any tasks that have been stopped
-        task_ids = session.exec(select(Task.task_id).where(*filter_query)).all()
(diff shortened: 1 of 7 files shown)


JarettForzano:

- #33
    - #34 ◀
        - #35
            - #36
                - #37
                    - #38
                        - #39

services/tracker/tests/unit/test_benchmark_utils.py
```diff
@@ -191,6 +192,36 @@ def get_test_session():
         benchmark_row = fetch_benchmark_row(benchmark_row.id, database_session)
         assert benchmark_row.status == BenchmarkStatus.IN_PROGRESS
 
+        # Reset benchmark row to stopped state
```
oalmatov, line 195: nit: now that i've had a break from the harness, i do understand Langstons point better. this should probably be name Job or smth like that. Benchmark is misleading imo, but can be done later.

services/tracker/src/tracker/utils.py
```diff
@@ -735,11 +743,18 @@ async def resume_benchmark(
...
-            .where(col(Task.benchmark) == benchmark_row.id)
-            .where(col(Task.status) == TaskStatus.STOPPED)
-            .values(status=TaskStatus.STARTING, started_at=datetime.now(ZoneInfo("UTC")))
+            .where(*filter_query)
+            .values(  # Reset to defaults
+                status=TaskStatus.STARTING,
+                started_at=datetime.now(ZoneInfo("UTC")),
+                error_message=None,
+                finished_at=None,
+            )
```
oalmatov, line 736: i think we should create new Task objects instead of updating them. I think it's useful information to see what's been failing. Also if we run with a --force in the future on tasks that have actually succeeded, we should keep track of the differences imo.

eventually we will probably be running benchmarks with pass@k and adding this logic now will make that implementation pretty easy.

services/tracker/tests/unit/test_benchmark_utils.py
JarettForzano: I saw that comment but I would like more of an explanation on why we should change it to job. This is the way I think about it, job is a more technical term used to talk about how we are running the benchmark and is not an accurate description of the object we are referencing and what we want the user to connect to.

When you run a benchmark we don't say,  "I am running a job", we say "I am running a benchmark". I think its fine in vcb since its not user facing so you can call it whatever you want. We are just introducing an additional learning curve to understanding and modifying the harness.

TLDR: Its not recommended to mix user facing paradigms with terms used to describe technical parts of an implementation. Even less so if we do not expect the user to know what a job is.

services/tracker/src/tracker/utils.py
JarettForzano: I wanted to stay away from versioning for two reasons (should avoid versioning if not nessecary)
1. Makes us build in a lot more architecture and is hard to use if cutoffs are not clearly defined
2. Its not really going to be used, i don't think I have ever seen this feature built in any harness. If you want keep the errors you can just download the results locally and store them somewhere. 

If you want to run a benchmark x times then we should create an additional layer of abstraction that groups the benchmarks. That way we are extending the harness instead of rebuilding it.

That is also more of an enhancement and I don't want to rebuild a portion of the harness to add in a feature thats just a nice to have.

oalmatov (approved):

PR #35, opened 2026-01-28
Resume --force flag feature

JarettForzano:
### Main changes
- Added a `--force` flag to the resume feature which allows you to specify a list of tasks that you want to retry while resuming a run
- Added unit test with the other resume unit tests
- Updated docs

### Testing
- You can just finish a single task in a run, stop the run then resume again using the --force flag to retry that task

### Additional Notes
- I am also going to make a separate cli to just retry a finished benchmark. Right now these flags are only good for if you want to resume a benchmark. The functionality is the same so its really just going to be an alias with a slight tweak to the utils
- Another thing I was thinking about is that maybe instead of `--force` we can do `--task_ids`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,7 @@ Flags
 
 ```
 --retry: Retry tasks with the status `error`
+--force task_1 task_2: force retry tasks with the specified space separated ids
 ```
 
 #### List and filter benchmarks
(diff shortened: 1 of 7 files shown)


JarettForzano:

- #33
    - #34
        - #35 ◀
            - #36
                - #37
                    - #38
                        - #39
                            - #40

oalmatov (approved):

PR #38, opened 2026-01-29
Update docs + references

JarettForzano:
### Main changes
- Update docs where contract was mentioned
- Updated code base to be more consistent with the terminology ex. `agent_name` instead of `contract_name`

PR #37, opened 2026-01-29
UV tool docs

JarettForzano:
### Main changes
- Added new make install-tool command
- Updated documentation to show how to install the tool
- Updated outdated doc

Turns out we already had the script in the pyproject.toml file, we just needed to document it

PR #36, opened 2026-01-29
Clean up logs

JarettForzano:
### Main changes
- Removed file handler (we don't want to dynamically increase the size of the docker container)
- Removed utility no longer in use

PR #40, opened 2026-02-02
Runtime bug fixes

JarettForzano:
### Main changes
- Broker does not timeout after 10 minutes
- Can resume a run with the status error
- Fixed bug where we were not fetching remaining tasks when evaluating after resuming
- Simplified how we stop a benchmark
- As a fall back i made it so that you can still stop a benchmark if it has an error state

### Description
This pr addresses a lot of the runtime errors I was running into and general usability issues a spotted

### DIFF
--- a/services/tracker/main.py
+++ b/services/tracker/main.py
@@ -263,8 +263,9 @@ async def stop_benchmark(
     if not benchmark_row:
         raise HTTPException(status_code=404, detail=f"Benchmark with id {benchmark_id} not found")
 
-    # Can only pause a in progress benchmark
-    if benchmark_row.status not in [BenchmarkStatus.IN_PROGRESS, BenchmarkStatus.ERROR]:
+    valid_stop_states = [BenchmarkStatus.IN_PROGRESS, BenchmarkStatus.ERROR, BenchmarkStatus.STOPPING]
+
+    if benchmark_row.status not in valid_stop_states:
         raise HTTPException(
             status_code=400,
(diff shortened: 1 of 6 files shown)


JarettForzano:

- #33
    - #34
        - #35
            - #36
                - #37
                    - #38
                        - #39
                            - #40 ◀

services/tracker/main.py
```diff
@@ -300,10 +300,10 @@ async def resume_benchmark(
     if not benchmark_row:
         raise HTTPException(status_code=404, detail=f"Benchmark with id {benchmark_id} not found")
 
-    if benchmark_row.status != BenchmarkStatus.STOPPED:
+    if benchmark_row.status not in [BenchmarkStatus.STOPPED, BenchmarkStatus.ERROR]:
```
OrestesK: make some var array with this and reuse

services/tracker/src/tracker/utils.py
```diff
@@ -343,6 +343,25 @@ def create_task_rows(
     return pending_task_rows
 
 
+async def fetch_missing_tasks(
```
OrestesK, line 346: good. one thing you can also do it put basic sql queries in their own function, like get_valid_tasks and then all code using that will be simpler

OrestesK (approved):

services/tracker/src/tracker/utils.py
JarettForzano: yeah, when i do a pass on the db queries i plan on making a utilities like that since it gets pretty verbose

PR #41, opened 2026-02-02
feat: setup aws infra

OrestesK:


infra/pyproject.toml
```diff
@@ -0,0 +1,27 @@
...
+]
+version = "0.1.0"
+description = "AWS CDK infrastructure for agentic harness services"
+readme = "README.md"
+requires-python = ">=3.12"
+
+dependencies = [
+    "aws-cdk-lib>=2.233.0",
+    "constructs>=10.4.4",
+]
```
JarettForzano, line 14: Do you think we should keep it separate or make a group inside of the top level pyproject.toml?

infra/Makefile
```diff
@@ -0,0 +1,46 @@
...
+install:
+	uv venv --python 3.12
+	uv sync
+	@echo "Done! Run 'source .venv/bin/activate' to activate."
+
+venv_check:
+	@if [ ! -f .venv/bin/activate ]; then \
+		echo "Virtualenv not found! Run 'make install' first."; \
+		exit 1; \
+	fi
```
JarettForzano, line 24: Should we make this a prereq for when we run the make commands below?

JarettForzano (approved):
mmmm, good work. not a priority but when you can ff to add CD for prod

infra/Makefile
OrestesK: yes, woops

PR #44, opened 2026-02-03
cloudwatch

JarettForzano:
cloudwatch

Updated daytona, cleaned up cloudwatch usage, refactored for streaming

Add a fixture to mock aws services in unit tests

Separate the mocked services

Fix integration tests

PR #43, opened 2026-02-03
commit error tasks to database

JarettForzano:
### Main changes
- Fix query bug
- Commit errors to database when we exit inside of the task tracker context

PR #50, opened 2026-02-05
Added retry feature

JarettForzano:
### Main changes
- Added an alias for resume-benchmark to retry-benchmark
- Adjusted endpoint to be runnable if the user has requested to retry a benchmark or resume it
- Updated tests

### Reasoning
The logic is the same so I was wondering what to do with the endpoint, I decided the best way to not confuse the user was to make it an alias.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,12 @@ Flags
 uv run harness resume-benchmark --benchmark-id <benchmark_id>
 ```
 
+#### Retry a benchmark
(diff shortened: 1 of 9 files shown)


JarettForzano:

- #42
    - #43
        - #44
            - #50 ◀
                - #51
                    - #52

OrestesK (approved):

PR #51, opened 2026-02-06
agent output change

JarettForzano:
### Main changes
- Removed agent_output db field
- Added support for creating a tar file
- Updated contract info


JarettForzano:

- #42
    - #43
        - #44
            - #50
                - #51 ◀
                    - #52

services/tracker/src/tracker/sandbox.py
```diff
@@ -181,12 +182,31 @@ async def stream_command_output(
             pass
 
 
+async def archive_and_upload_output(sandbox: AsyncSandbox, output_path: str, agent_output_s3_key: str) -> None:
+    """Compress a file in the sandbox into a tar.gz and upload it to S3"""
+    archive_path = f"/tmp/{uuid.uuid4().hex}.tar.gz"
+
+    tar_result = await sandbox.process.exec(f"tar -czf {shlex.quote(archive_path)} {shlex.quote(output_path)}")
```
OrestesK, line 189: is there no way to use the native tar python package? clean would be make a function in the sandbox called archive_outputs and then gRPC call that.

OrestesK (approved):

services/tracker/src/tracker/sandbox.py
JarettForzano: i think just about all linux containers come with tar downloaded as default which is why we are using that command

PR #53, opened 2026-02-10
fix: dev and logging

OrestesK:


services/tracker/src/tracker/sandbox.py
```diff
@@ -162,12 +218,20 @@ async def stream_command_output(
...
+        try:
+            await sandbox.process.get_session_command_logs_async(
+                session_id=session_id,
+                command_id=cmd_id,
+                on_stdout=on_output,
+                on_stderr=on_output,
+            )
+        except DaytonaError as e:
+            error_msg = str(e).lower()
+            if "ping timeout" in error_msg or "websocket" in error_msg:
```
JarettForzano: If you reverse the statement you can get rid of the else

if not websocket issue:
     raise ....

await _poll_.....

services/tracker/src/tracker/sandbox.py
```diff
@@ -141,13 +160,50 @@ async def install_agent_dependencies(
...
+        # Fetch current logs
+        try:
+            logs = await sandbox.process.get_session_command_logs(session_id, cmd_id)
+            current_output = logs.output or ""
+
+            # Output only new content
+            if len(current_output) > last_log_length:
+                new_content = current_output[last_log_length:]
+                on_output(new_content)
+                last_log_length = len(current_output)
```
JarettForzano, line 189: Did the buffering at the top not work?

```
log_queue: asyncio.Queue[str] = asyncio.Queue(maxsize=20)

# Collects the logs and dumps them when the queue is full
def log_output(data: str) -> None:
    log_queue.put_nowait(data)
    buffer_logs(log_queue, stream_key)
```

services/tracker/src/tracker/sandbox.py
```diff
@@ -28,6 +31,22 @@
...
+"""
+
+_original_connect = ws_client.connect
+
+
+def _patched_connect(*args: object, **kwargs: object):
+    kwargs.setdefault("ping_interval", 60)
+    kwargs.setdefault("ping_timeout", 120)
+    kwargs.setdefault("close_timeout", 30)
+    return _original_connect(*args, **kwargs)  # pyright: ignore[reportArgumentType]
```
JarettForzano, line 46: sexy, pretty sure this like quadrupled the time till we got the timeout but the error still occurs. polling works pretty good so you could probably remove this since its not a perm fix. up2u

services/tracker/src/tracker/database/models.py
```diff
@@ -30,20 +30,20 @@
...
+    FINISHED = "FINISHED"
+    ERROR = "ERROR"
 
 
 class BenchmarkStatus(str, Enum):
-    IN_PROGRESS = "in_progress"
-    STOPPING = "stopping"
-    STOPPED = "stopped"
-    FINISHED = "finished"
-    ERROR = "error"
```
JarettForzano, line 46: This broke some assumptions we are making about the string format of the status. might be worth reverting or fixing the cli bugs

services/tracker/src/tracker/sandbox.py
OrestesK: this only does it every 5 seconds so it's fine

src/agentic_harness/cli/utils.py
```diff
@@ -17,28 +17,14 @@
...
-            "stopped": "cyan",
-            "finished": "green",
-            "error": "red",
-        }
-        return status_colors.get(status_value, "white")
+    STATUS_COLORS = {
+        "PENDING": "yellow",
+        "IN_PROGRESS": "blue",
+        "EVALUATING": "magenta",
+        "STOPPED": "cyan",
```
JarettForzano, line 24: Missing `STOPPING` status, would cause a runtime error. I suggest using a `.get` with a default value in case someone adds a new status in the future / modifies one and we have a bruh moment in production

JarettForzano (approved):
Good to merge after fixing bug

src/agentic_harness/cli/utils.py
OrestesK: it is actually a very good practice to never do .get(). Usually with ENUMS, you even want a match, where the default is to raise an exception. This makes it so when someone adds a new ENUM, they are forced to handle all cases where it is used. Adding STOPPING

JarettForzano: hmm, i agree but ill still make a test then so that we don't create any runtime errors

JarettForzano (approved):

PR #54, opened 2026-02-10
Download agent outputs

JarettForzano:
### Main changes
- Added two knew endpoints
   - fetch-agent-outputs: Returns a streaming response that streams the tar file to the client where we download it to the folder specified by the user
   - fetch-benchmark-metadata: Returns metadata about the benchmark to the client, made to be small and include metadata about the file to ensure a user can still find it when they do not specify the name themselves
- Updated documentation to include new fetch-agent-outputs command arg
- Modified the entrypoint script, there was a bug with using wait to wait until the database endpoint was reachable (we also already use the health endpoint so it was redundant)
- Added a yieldingwriter class, we inherit the base class and override the ones necessary to stream bytes to the client
- Added a list s3 objects and download stream utilities to assist with downloading tar files 

### Reasoning
We need to save agent outputs to s3 so we also need to provide a way to retrieve them

### DIFF
--- a/services/tracker/entrypoint.sh
+++ b/services/tracker/entrypoint.sh
@@ -1,9 +1,6 @@
 #!/bin/bash
 set -e
 
-echo "Waiting for database..."
-dockerize -wait tcp://${DB_HOST}:${DB_PORT} -timeout 120s
-
 echo "Running database migrations..."
 uv run alembic upgrade head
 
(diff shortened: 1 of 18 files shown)


JarettForzano:

- #56
    - #54 ◀
        - #58

services/tracker/entrypoint.sh
```diff
@@ -1,9 +1,6 @@
 #!/bin/bash
 set -e
 
-echo "Waiting for database..."
-dockerize -wait tcp://${DB_HOST}:${DB_PORT} -timeout 120s
-
```
JarettForzano, line 6: this was not working

PR #71, opened 2026-02-14
Bug fixes

JarettForzano:
### Main changes
- Use path instead of text
- Handle no tasks returned

PR #72, opened 2026-02-14
s3 retrieve-results feature

JarettForzano:
### Main changes
- Allow uploading results to s3 instead of downloading to disk
- create presigned url
- create url to s3 bucket dashboard
- union for two different return types now (s3 or just the final view)
- endpoint for checking if file exists in s3 already
- updated documentation


JarettForzano:

- #71
    - #72 ◀
        - #73

PR #121, opened 2026-02-25
add CLAUDE.md

vatsj:


oalmatov (approved):
Not sure why claude can't just read the README but all good

PR #146, opened 2026-03-06
add dummy agent

vatsj:


services/tracker/src/tracker/sandbox.py
```diff
@@ -180,6 +179,8 @@ async def upload_agent_artifacts(
 
     if files_to_upload:
         await sandbox.fs.upload_files(files_to_upload)
+    else:
+        await sandbox.fs.create_folder(str(bundle_path / contract.name), "755")
```
JarettForzano, line 183: Why do we need to create a folder?

vatsj: cuz we run the agent from that folder. Specifically sandbox.py:199 fails:
> await stream_command_output(sandbox, f"**cd {str(contract_path)}** && {contract.install_cmd}", log_output)

JarettForzano: it might be better to adjust the contract path instead of making an empty folder but i'll approve for now

JarettForzano (approved):

PR #148, opened 2026-03-06
add: recalculate excel tool

vatsj:


JarettForzano:
Deprecated, should be moved to model registry

PR #159, opened 2026-03-10
taskmonitor cancels running tasks when force stop

Chen-Oliver:
TaskMonitor should track all tasks until they reach the DONE state, so that if we force stop the benchmark it can cancel the tasks directly. Currently, force stop can take a while to actually reach the benchmark STOPPED state because we no longer have a handle to the task and can't cancel them if they have already acquired the semaphore.

### DIFF
--- a/services/tracker/src/tracker/database/models.py
+++ b/services/tracker/src/tracker/database/models.py
@@ -271,9 +271,13 @@ class Task(SQLModel, table=True):
     @computed_field
     @property
     def alias(self) -> str:
-        """Unique alias for the current task attempt, used when creating sandboxes."""
+        """Unique alias for the current task attempt, used when creating sandboxes.
+
+        Format: {task_id}_{attempt_suffix}
+        - attempt_suffix: hex-encoded microsecond timestamp from started_at, changes on each retry
+        """
         attempt_suffix = f"{int(self.started_at.timestamp() * 1_000_000):x}"
-        return f"{self.task_id}_{self.id.hex[:5]}_{attempt_suffix}"
+        return f"{self.task_id}_{attempt_suffix}"
 
 
 @event.listens_for(Task, "before_insert")
--- a/services/tracker/src/tracker/database/models.py
+++ b/services/tracker/src/tracker/database/models.py
@@ -273,11 +273,11 @@ class Task(SQLModel, table=True):
     def alias(self) -> str:
         """Unique alias for the current task attempt, used when creating sandboxes.
 
-        Format: {task_id}_{attempt_suffix}
-        - attempt_suffix: hex-encoded microsecond timestamp from started_at, changes on each retry
+        Format: {task_id}_{suffix}
+        - suffix: hex-encoded microsecond timestamp from started_at, changes on each retry/resume
         """
-        attempt_suffix = f"{int(self.started_at.timestamp() * 1_000_000):x}"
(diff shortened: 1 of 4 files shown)


services/tracker/src/tracker/utils.py
```diff
@@ -186,29 +193,18 @@ def _validate_task(self, task_id: str) -> bool:
             benchmark_row = fetch_benchmark_row(self._benchmark_id, session)
             task_row = self._fetch_task_row(task_id)
 
-            # If task has been stopped or benchmark has occured an error we need to exit
-            if task_row.status == TaskStatus.STOPPED or benchmark_row.status == BenchmarkStatus.ERROR:
+            # If task has been stopped or benchmark is stopping / errored we need to exit
+            if task_row.status == TaskStatus.STOPPED or benchmark_row.status in [
+                BenchmarkStatus.ERROR,
```
JarettForzano: Would adding the `BenchmarkStatus.STOPPING` also affect stopping the benchmark without the --force flag?

services/tracker/src/tracker/utils.py
```diff
@@ -135,6 +135,15 @@ async def _wrap_coro():
...
         except asyncio.CancelledError:
             logger.warning(f"Task {task_row.task_id} was cancelled")
+            with Session(bind=engine) as session:
+                task = fetch_task_row(task_row.id, session)
+                benchmark = fetch_benchmark_row(task.benchmark, session)
+
+                if benchmark.status == BenchmarkStatus.STOPPING or task.status == TaskStatus.STOPPED:
+                    task.status = TaskStatus.STOPPED
+                    session.add(task)
+                    session.commit()
```
JarettForzano: Would this be a double commit (if task status is stopped set it to stopped)? We set the task status to stopped inside of the stop benchmark endpoint, the tasks set to stopped are cancelled causing the `asyncio.CancelledError` which is expected, I think at this point the status is already equal to `TaskStatus.STOPPED`.

services/tracker/src/tracker/database/models.py
```diff
@@ -271,8 +271,13 @@ class Task(SQLModel, table=True):
...
     def alias(self) -> str:
-        """Unique alias for the task that is used to uniquely identify the same task when creating sandboxes"""
-        return f"{self.task_id}_{self.id.hex[:5]}"
+        """Unique alias for the current task attempt, used when creating sandboxes.
+
+        Format: {task_id}_{attempt_suffix}
+        - attempt_suffix: hex-encoded microsecond timestamp from started_at, changes on each retry
+        """
+        attempt_suffix = f"{int(self.started_at.timestamp() * 1_000_000):x}"
+        return f"{self.task_id}_{attempt_suffix}"
```
JarettForzano: Good solution, its not really an attempt though since it cannot fail but thats fine

services/tracker/src/tracker/utils.py
Chen-Oliver: yeah you're right, this would also kill running tasks in the non-force stops. can just rely on TaskStatus.STOPPED check directly since non-force doesn't set IN_PROGRESS or EVALUATING tasks to stopped statsu.

services/tracker/src/tracker/utils.py
Chen-Oliver: yep good catch

JarettForzano (approved):
Approved

PR #166, opened 2026-03-11
Integrate OpenHands VCB runtime with tracker sandbox improvements

tthuwng:
## Summary
- improve the OpenHands runtime path used for VCB app-generation runs
- add generic snapshot-backed Daytona sandbox support and better tracker cancellation/logging
- keep local agent-path development compatible with the newer harness workflow

## Included changes
- package full OpenHands outputs under `/tmp/openhands_output`
- run OpenHands against the full VCB prompt/workspace contract
- support `snapshot:<name>` sandbox sources in tracker
- improve task cancellation for tasks already past the queueing stage
- reduce noisy stream-log spam and preserve streamed runtime lines better


services/tracker/src/tracker/utils.py
```diff
@@ -235,6 +234,11 @@ def handle_early_exit(task_row: Task, task_session: Session) -> None:
     task_session.commit()
 
 
+def create_sandbox_name(task_row: Task) -> str:
+    """Use a per-attempt suffix so retries/resumes do not collide with stale sandbox names."""
+    return f"{task_row.alias}_{uuid4().hex[:5]}"
```
Chen-Oliver: task_row.alias is already unique per retry/resume so this isn't necessary

tests/test_openhands_wrapper.py
JarettForzano, line 1: ~I don't think we should be having tests like this at the top level, the benchmark service should have the tests that ensures that it works. i think the tracker just needs to connect to it~

Realizing this was a test for the agent, but same comment applies. We should not have tests for agents inside of valk

services/tracker/src/tracker/utils.py
```diff
@@ -235,6 +234,11 @@ def handle_early_exit(task_row: Task, task_session: Session) -> None:
     task_session.commit()
 
 
+def create_sandbox_name(task_row: Task) -> str:
+    """Use the task alias, which is already unique for each retry/resume attempt."""
+    return task_row.alias
```
JarettForzano, line 239: Utility not needed

agents/openhands/base_openhands_config.toml
JarettForzano, line 1: Move agent to https://github.com/vals-ai/agent-registry

PR #167, opened 2026-03-12
Added timeout to agent run command + Updated claude code contract

JarettForzano:


agents/claude_code/contract.py
```diff
@@ -56,7 +56,9 @@ def run_cmd(self, problem_statement_path: str, task_id: str, kwargs: dict[str, A
             f"--allowedTools {' '.join(self._ALLOWED_TOOLS)}",
         ]
 
-        run_cmd = f"cat {problem_statement_path} | claude " + " ".join(args) + " 2>&1 | tee /logs/claude_code.log"
+        run_cmd = (
+            f"cat {problem_statement_path} | stdbuf -oL claude " + " ".join(args) + " 2>&1 | tee /logs/claude_code.log"
+        )
```
JarettForzano, line 61: View the logs inside of cloudwatch while streaming the text to a log

services/tracker/src/tracker/sandbox.py
```diff
@@ -311,6 +314,13 @@ async def run_agent(
 
     run_cmd = contract.run_cmd.replace("{problem_statement_path}", problem_path).replace("{task_id}", task_id)
 
+    # Apply timeout if specified
+    if agent_timeout is not None:
+        run_cmd = f"timeout {agent_timeout} {run_cmd}"
+
+    # Create cwd if it does not already exist
+    await sandbox.process.exec(f"mkdir -p {shlex.quote(cwd)}")
```
JarettForzano, line 322: Recognized this needed to be added, if the benchmark author expected the agent to be executed here we should just ensure it does

src/agentic_harness/cli/main.py
```diff
@@ -464,8 +464,10 @@ def start(
         if agent_path.is_dir():
             asyncio.run(push_agent(agent_path.stem, agent_path))
             contract = get_contract(agent_path / "contract.py", agent_config)
+            contract.name = agent_path.stem
         else:
             contract = asyncio.run(get_contract_from_s3(agent, agent_config))
+            contract.name = agent
```
JarettForzano, line 470: If the user pulls the agent from S3 we have to make sure that the contract name is the same as where the bundle path is in S3. Issue was created for this since it is weird

PR #200, opened 2026-03-19
add Slack webhook notifications for benchmark runs

### LINKED ISSUE #107: Add an option for a slack webhook that provides notifications when runs complete
# Slack Webhook Notifications for Benchmark Runs

## Dev setup 

Follow these steps to setup the dev environment where you will be testing this feature.

1. Navigate to the [dev docs ](https://github.com/vals-ai/Valkyrie/blob/dev/docs/DEVELOPMENT.md). Follow all the steps but the section on `Submodules`. 
2. Navigate to the [benchmark-registry](https://github.com/vals-ai/benchmark-services-registry). Using the [make commands section](https://github.com/vals-ai/benchmark-services-registry?tab=readme-ov-file#commands), initialize all the submodules for the project
3. [ Install dependencies for ngrok ](https://github.com/vals-ai/create-benchmark-service?tab=readme-ov-file#reverse-tunnel-setup)
4. cd into `terminal-bench-benchmark-service`, this is how i run it with ngrok (slight deviation from readme). `make docker-build`; `docker run -d -p 8001:8001 benchmark-service:latest` (running it in a detached state), `ngrok http 8001`
5. Navigate back to the valk cli and set the ngrok url as [a custom benchmark service](https://github.com/vals-ai/Valkyrie?tab=readme-ov-file#set-a-custom-benchmark-service), overriding the default one we have setup. Use the _exact_ url, no additional slashes at the end.

After you have completed these steps you will have the cli pointing to the dev version of the tracker service you started using `make tracker-service` by setting the env `TRACKER_SERVICE_URL=http://localhost:8000` in the root of your project, a reverse tunnel routing the proxy setup from the ngrok url to the benchmark service running locally on your machine. If you have trouble contact @JarettForzano 

Example run command to test that it is working `valkyrie run start --agent codex --model gpt-5.3-codex --benchmark terminal-bench --concurrency 10 --slice :1` (don't worry the agent is already uploaded to the s3 bucket `agentic-harness`

## Overview

Add support for Slack webhook notifications during benchmark runs. Users can configure a webhook URL and define progress intervals at which notifications are sent, making it easy to monitor multiple concurrent runs without actively watching the CLI.

This feature is modeled after the existing `--connect` flag behavior in `valk run fetch`, extended to support push notifications.

---

## Proposed Behavior

When starting a benchmark run, users can optionally provide a Slack webhook URL and one or more progress intervals. Notifications are sent automatically as the run progresses at those thresholds. We should always send a notification when a benchmark reaches a finished status (e.g., finished, error, stopped)

### Starting a Run with Notifications

```bash
valk run start .... -i 25 -i 75 -i 100
```

- `-i` / `--interval` defines the percentage threshold at which a notification is sent
- The webhook URL is sourced from the user's config (created using `valk config init`). You will need to add a new env var inside that we can soruce.

---

## Notification Triggers & Format

Notifications are sent on the following conditions:

| Trigger | Description |
|---|---|
| **In Progress** | Run has crossed a defined interval threshold and is still ongoing |
| **Error** | Run has errored out and an `error_message` has been set |
| **Finished** | All tasks within the benchmark have completed |
| **Stopped** | The user has requested a stop and the status is set to `STOPPED` |

Each notification includes only the base run information and percent completed. **Accuracy is excluded from in-progress notifications** — it is only calculated at the end of a run.

> **Schema TBD** — please review the database objects and propose a notification payload schema before implementation so we can align on the structure. Tag @JarettForzano with your proposal.

---

### Webhook URL

- Store the webhook URL in the user's YAML config via `valk config`
- When a run starts, the webhook URL should also be persisted to the benchmarks database object so it can be reused if the run is stopped and later resumed — the user should only need to provide it once at runtime either through a config or cli.

### Interval Restrictions

To keep notification volume manageable, the following constraints apply to `-i`:

- **Maximum of 3 intervals** per run
- Each value **must be divisible by 5**
- Valid range: **1–100**
- Example: `-i 25 -i 75 -i 100`

---

## Testing

- Include a **screen recording** in the PR description demonstrating the feature end-to-end
- This is classified as an **enhancement** — integration tests are not required if they would be superficial. No tests are preferred over filler tests

---

## Open Questions

- [ ] Notification payload schema — needs to be defined and agreed upon before implementation (tag @JarettForzano)
- [ ] Should we just do a single source of truth for the webhook url and keep it inside of the config or also allow using the cli?
- [ ] Should we make the default if not specified -i 100 if webhook url exists?

BradenBug:
## Summary
Add support for sending Slack notifications at user-defined progress
  thresholds during benchmark runs. Users configure a webhook URL via
  `valkyrie config webhook set` and specify intervals with `-i` on
  `valk run start`.

## Changes
  - SlackNotifier with threshold tracking and fire-and-forget delivery
  - Progress notifications checked in TaskMonitor polling loop
  - Terminal notifications on finish/error/stop with final score
  - Interval validation (max 3, divisible by 5, range 5-100)
  - Default to 100% notification when webhook URL is configured
  - Webhook URL excluded from X-Harness headers
  - Resume/retry automatically re-reads webhook URL from config

## Related Issues
Closes #107 

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor
- [ ] Docs / config

## Testing
- [x] Manually tested

## Checklist

- [x] Self-reviewed the diff
- [x] Docs updated if needed

## Future Improvements
- Slack message styling could be improved
- Include S3, Cloudwatch links in slack messages (would probably require a refactor at this point)
- Persist webhook details in DB for consistency in resumes between users
- Revisit cli config pattern
- Create object to hold harness config to reduce the amount of static get methods inside of the tracker context manager

## Demo

https://github.com/user-attachments/assets/6edff410-abb1-4bc6-91a6-6f68474ea9c9


.vscode/settings.json
JarettForzano, line 1: Didn't like the workspace vscode setting?

services/tracker/src/tracker/notifications.py
```diff
@@ -0,0 +1,169 @@
...
+    from tracker.types import BenchmarkDetails
+
+logger = get_logger(__name__)
+
+WEBHOOK_TIMEOUT = 5.0
+
+
+@dataclass(frozen=True)
+class NotificationContext:
+    """Common fields for all notification messages."""
```
JarettForzano: What was the reasoning behind using a dataclass instead of a pydantic model? I would say to stick with pydantics to keep the code consistent

services/tracker/src/tracker/notifications.py
```diff
@@ -0,0 +1,169 @@
...
+            benchmark_name=benchmark_row.name,
+            agent_name=benchmark_row.arguments.contract.name,
+            benchmark_id=benchmark_row.id,
+            started_at=benchmark_row.started_at,
+            total_tasks=details.total_tasks,
+            finished_tasks=details.finished_tasks,
+        )
+
+
+def _format_duration(started_at: datetime) -> str:
```
JarettForzano, line 51: I would try to structure this to be less noisy

```python
def _format_duration(started_at: datetime) -> str:
    """Format elapsed time as human-readable string."""
    if started_at.tzinfo is None:
        started_at = started_at.replace(tzinfo=timezone.utc)

    seconds = int((datetime.now(timezone.utc) - started_at).total_seconds())

    if seconds < 60:
        return f"{seconds}s"

    minutes, _ = divmod(seconds, 60)
    if minutes < 60:
        return f"{minutes}m"

    hours, minutes = divmod(minutes, 60)
    return f"{hours}h {minutes}m"
```

services/tracker/src/tracker/notifications.py
```diff
@@ -0,0 +1,169 @@
...
+
+    async def __aenter__(self) -> "SlackNotifier":
+        return self
+
+    async def __aexit__(self, *_: object) -> None:
+        await self.close()
+
+    async def close(self) -> None:
+        """Close the HTTP client."""
+        await self._client.aclose()
```
JarettForzano: When it comes to context managers aim to use the lib already exposed

https://docs.python.org/3/library/contextlib.html#contextlib.asynccontextmanager

JarettForzano: Does not look like its being used like a context manager but the dunder methods suggest the intention

services/tracker/src/tracker/utils.py
```diff
@@ -195,6 +200,19 @@ def _validate_task(self, task_id: str) -> bool:
 
         return True
 
+    async def _check_notifications(self) -> None:
+        """Check notification thresholds using DB task counts."""
+        if not self._notifier:
+            logger.info("No notifier")
```
JarettForzano: This log has the potential to get noisy considering the # of cycles we are going to call the check notif method for. I would remove it since its just a guard

services/tracker/src/tracker/utils.py
```diff
@@ -599,6 +631,16 @@ async def process_benchmark(
 
             set_benchmark_final_status(benchmark_row, session)
 
+            # Send terminal notification
+            if notifier:
```
JarettForzano: To reduce code smells between sections 634-642 and 660-673 you can just call this inside of the final statement. Status will be updated at that point

services/tracker/src/tracker/notifications.py
```diff
@@ -0,0 +1,169 @@
...
+
+    benchmark_name: str
+    agent_name: str
+    benchmark_id: UUID
+    started_at: datetime
+    total_tasks: int
+    finished_tasks: int
+
+    @classmethod
+    def from_benchmark(cls, benchmark_row: Benchmark, details: BenchmarkDetails) -> NotificationContext:
```
JarettForzano: `BenchmarkDetails` can be retrieved from a `Benchmark` row, you can simplify this repetitive pattern

```
benchmark_row = fetch_benchmark_row(self._benchmark_id, session)
benchmark_details = BenchmarkContext(benchmark_row, session).benchmark_details
```

services/tracker/src/tracker/notifications.py
```diff
@@ -0,0 +1,169 @@
...
+        self._fired: set[int] = set()
+        self._client = httpx.AsyncClient(timeout=WEBHOOK_TIMEOUT)
+
+    async def _send_webhook(self, text: str) -> None:
+        """Fire and forget — exceptions are caught and logged, never raised."""
+        try:
+            response = await self._client.post(
+                self._webhook_url,
+                json={"text": text},
+            )
```
JarettForzano: You can use the context manager the client provides

```
async with httpx.AsyncClient(timeout=WEBHOOK_TIMEOUT) as client:
    await client.post(...)
```

this removes the need to have  `async def close(self) -> None: ...` and manually close the client later

```
if notifier:
    await notifier.close()
```

src/valkyrie/cli/main.py
```diff
@@ -301,6 +302,56 @@ def auth_list() -> None:
         click.echo(f"  {name}: {masked}")
 
 
+@config.group()
+def webhook() -> None:
+    """Manage Slack webhook URL for run notifications."""
+    pass
```
JarettForzano, line 308: Tricky situation, do we want to create groups for each config like setting or make the current set of config options more flexible?

We currently expose a `valkyrie config modify <KEY> <VALUE>` command to modify a single key but we could change that to a `set` command that upserts a key (if we support it). This would only be viable for when we can map the intention with a value (e.g., timeout 360, we know its _timout_ they want set). For values where they belong to a _group_ like benchmark services, it may make more sense to have their own cli group. This is because if we asked someone to use the current schema they would have no choice but to use `<Service> <URL>`, which does not make it clear its a _service_ that they are setting since we have no indicator.

When I saw `valkyrie config webhook set <VALUE>` I realized that there is another paradigm we can use that avoids having to make a new `set` and `remove` command. An example of how I would map the current command would be `valkyrie config set webhook <VALUE>`, `webhook` is the `<KEY>`

lmk if that makes sense, something to think about since it allows us to reduce repetitive code.

src/valkyrie/cli/main.py
```diff
@@ -444,6 +504,25 @@ def start(
...
+        click.echo(
+            click.style(
+                "  Warning: --interval specified but no webhook URL configured. Run `valkyrie config webhook set <url>` first. Ignoring intervals.",
+                fg="yellow",
+            )
+        )
+    elif intervals:
+        webhook_intervals = validate_intervals(intervals)
+    elif webhook_url:
+        webhook_intervals = [100]
```
JarettForzano: You can set a default value using click

src/valkyrie/cli/main.py
```diff
@@ -694,13 +775,24 @@ def resume(
     if ctx.info_name == "retry":
         retry = True
 
+    # Read webhook URL from config for notifications
+    webhook_url = TrackerService.get_webhook_url()
+    webhook_intervals: list[int] | None = [100] if webhook_url else None
+
```
JarettForzano, line 768: I think we want to make these persist and store the information inside of the db, that way if a different person resumes the benchmark than the one who started it they won't need to have the same webhook information which is what we want to avoid (since something changes and they would have no idea what).

In relation we might want to have the user define a secret name that holds the webhook inside the secret manager, that way they can change the secret if they want mid run and we wont be storing sensitive information that gets exposed when the benchmark object is exported to a json file.

src/valkyrie/cli/tracker_service.py
```diff
@@ -113,6 +113,24 @@ def get_benchmark_auth(benchmark_name: str) -> str | None:
         auth = harness_config.get("benchmark_auth") or {}
         return auth.get(benchmark_name)
 
+    @staticmethod
+    def get_webhook_url() -> str | None:
```
JarettForzano, line 117: Git issue: Create object to hold harness config to reduce the amount of static get methods inside of the tracker context manager

src/valkyrie/cli/main.py
JarettForzano: Optional: Additional note is that lines 507-521 just validate and assign defaults, you can include these inside of a helper method to reduce the amount of processing thats at the top

JarettForzano (changes requested):
Good first pr into valk, if i misunderstood something you wrote lmk inside of the comment thread.

.vscode/settings.json
BradenBug: My bad, I thought this was just generated by my vscode, didn't realize there was one already there

services/tracker/src/tracker/notifications.py
BradenBug: Just unfamiliar with the conventions here. Will refactor.

services/tracker/src/tracker/notifications.py
BradenBug: Sure

services/tracker/src/tracker/notifications.py
BradenBug: Will do

services/tracker/src/tracker/utils.py
BradenBug: Yep

services/tracker/src/tracker/utils.py
BradenBug: Good idea

src/valkyrie/cli/main.py
BradenBug: Interesting, I didn't think of that. I'll see what I can do

src/valkyrie/cli/main.py
BradenBug: I think we can create a separate issue for this refactor

src/valkyrie/cli/main.py
BradenBug: Since this is a bigger design change I think I'll also defer this to a separate followup PR. Don't want to add too much context here.

src/valkyrie/cli/tracker_service.py
BradenBug: Do you have an agent hook to create issues from comments like this or are you just noting it here?

JarettForzano: Just noting it, i created a ticket for it already

src/valkyrie/cli/main.py
JarettForzano: Yes, https://github.com/vals-ai/Valkyrie/issues/203

src/valkyrie/cli/main.py
JarettForzano: Makes sense, created a ticket https://github.com/vals-ai/Valkyrie/issues/204

JarettForzano (approved):

PR #209, opened 2026-03-22
Persist model name to DB and expose in discovery

Chen-Oliver:
## Summary

Model name is not currently persisted in the database which makes distinguishing runs difficult.
## Changes

- Include model name as part of the agent contract args saved to db. expose model name when listing runs and allow filtering on it.
- fixed a bug in the runs list filtering where the filter for agent_name was broken due to `func.json_extract(col, "$.contract.name")` being sqlite-only syntax. 
Screenshot 2026-03-21 at 11 03 34 PM

## Related Issues

Closes #

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor
- [ ] Docs / config

## Testing

- [x] Manually tested

## Checklist

- [x] Self-reviewed the diff
- [x] Docs updated if needed


JarettForzano (approved):
Approved

PR #217, opened 2026-03-24
Make Daytona log streaming best-effort

tthuwng:
## Summary
Stop treating Daytona log-stream failures as task failures.

Tracker now:
1. starts the sandbox command
2. tries to stream logs live
3. retries websocket log-stream failures a few times in the background
4. keeps polling command status until completion
5. decides success from the command exit code

If log streaming still fails, tracker logs a warning and continues without live logs.

## Why
Hosted VCB runs hit errors like:

`Failed to get session command logs: server rejected WebSocket connection: HTTP 502`

That is a log transport failure, not a command failure. Tasks should not fail just because live logs are unavailable.


JarettForzano (changes requested):
Waiting on validation this actually solves the issues

JarettForzano (approved):
Approved

services/tracker/tests/unit/test_stream_command_output.py
```diff
@@ -64,52 +64,97 @@ class FakeSandbox:
...
-async def test_stream_command_output_raises_non_websocket_error() -> None:
-    @dataclass
-    class FailingProcess(FakeProcess):
-        async def get_session_command_logs_async(  # type: ignore[reportIncompatibleMethodOverride]
-            self, **kwargs: Any
-        ) -> None:
-            raise DaytonaError("some other error")
 
-    sandbox = FakeSandbox(process=FailingProcess())
+@pytest.mark.asyncio
```
JarettForzano: These don't need to be marked since we are using the pyproject.toml setting

PR #222, opened 2026-03-26
First pass deprecating modify for set

### LINKED ISSUE #203: Change `modify` config command to `set` and move webhook command inside
Reference comment
https://github.com/vals-ai/Valkyrie/pull/200#discussion_r2963163944

JarettForzano:
## Summary
As previously discussed a better paradigm for interacting with the config is through `set` rather than `modify`. `set` implies upsert while `modify` is only for changing. Also deprecated the webook group as it overlaps with this functionality. Setting the webhook is trivially close to the original command. We are using the same config key as before `slack_webhook_url` so no parsing logic will need to be changed.

```bash 
valk config set webhook xyz
```

### CLI Command images
**Setting a key that does not exist inside of the config values**
image

**Setting webhook**
```bash
valk config set webhook xy
```
**Key value pair set inside of config**
```yaml
slack_webhook_url: xyz
```

**Removing a key that is optional vs a required key**

```bash
jarettforzano@Jaretts-MacBook-Pro agentic-harness % valk config remove webhook
  webhook removed.
jarettforzano@Jaretts-MacBook-Pro agentic-harness % valk config remove LOG_GROUP
Error: Key 'LOG_GROUP' is required and cannot be removed. Consider using `valkyrie config set` to update it.
```

## Changes
- Changed modify to set
- Created a `class ConfigValue` enum which contains an exhaustive list of 1-1 mappings that exist for the set and remove command
- Deprecated webhook cli group
- Added remove command to the config group
- Pulled out environment variables required by `valk config init` to a constant that can be reused

## Related Issues

Closes https://github.com/vals-ai/Valkyrie/issues/203

## Type of Change

- [ ] Bug fix
- [ X ] New feature
- [ ] Breaking change
- [ X ] Refactor
- [ X ] Docs / config

## Testing

- [ X ] Manually tested

## Checklist

- [X  ] Self-reviewed the diff
- [X ] No debug/dead code left in
- [ X] Docs updated if needed

### DIFF
--- a/src/valkyrie/cli/utils.py
+++ b/src/valkyrie/cli/utils.py
@@ -6,6 +6,7 @@ import shutil
 import tarfile
 import tempfile
 from datetime import datetime
+from enum import Enum
 from pathlib import Path
 from typing import Any, Coroutine, TypeVar
 from uuid import UUID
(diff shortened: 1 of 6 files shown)


src/valkyrie/cli/utils.py
```diff
@@ -30,6 +31,32 @@
...
+    DAYTONA_SECRET_NAME = "DAYTONA_SECRET_NAME"
+    LOG_GROUP = "LOG_GROUP"
+    LOG_RETENTION_POLICY = "LOG_RETENTION_POLICY"
+
+    @classmethod
+    def from_str(cls, key: str) -> "ConfigValue":
+        """Convert string value to enum value, raising if value is not an option."""
+        normalized = key.lower()
+        # Extra mapping for aliases, to be used for backwards compatibility
+        aliases = {"webhook": cls.SLACK_WEBHOOK_URL}
```
BradenBug: I don't think this is necessary, I don't think anyone has built workflows off of the existing command yet

BradenBug (approved):
LGTM!

src/valkyrie/cli/utils.py
JarettForzano: fair, ill remove the backwards compatibility support. can you change your config manually?

BradenBug: yep

PR #237, opened 2026-03-31
Add retry logic for sandbox exec commands

oalmatov:
## Summary
- Added `exec_with_retries` helper that retries `sandbox.process.exec()` calls up to 3 times on `DaytonaError`
- Applied to `mkdir -p` and `test -e` commands in `run_agent` which were failing intermittently (206 task errors in a recent fab benchmark run)
- Added unit tests for the retry logic

## Test plan
- [x] Unit tests pass (`pytest tests/unit/test_stream_command_output.py`)


services/tracker/src/tracker/sandbox.py
```diff
@@ -353,6 +355,21 @@ async def archive_and_upload_output(
...
+        try:
+            return await sandbox.process.exec(command)
+        except DaytonaError:
+            if attempt == max_attempts - 1:
+                raise
+
+            logger.warning(f"exec failed (attempt {attempt + 1}/{max_attempts}), retrying: {command}")
+            await asyncio.sleep(1)
+
+    raise RuntimeError("Unreachable")
```
tthuwng: this is very sus

services/tracker/src/tracker/sandbox.py
```diff
@@ -353,6 +355,21 @@ async def archive_and_upload_output(
             logger.warning(f"File {archive_path} does not exist, skipping removal")
 
 
+async def exec_with_retries(sandbox: AsyncSandbox, command: str, max_attempts: int = 3) -> Any:
+    """Execute a command in the sandbox, retrying on DaytonaError."""
+    for attempt in range(max_attempts):
+        try:
+            return await sandbox.process.exec(command)
+        except DaytonaError:
```
BradenBug: Have you confirmed process.exec raises a DaytonaError? Docs might be wrong but they don't indicate so:
https://www.daytona.io/docs/en/python-sdk/sync/process/#processexec

oalmatov: They do, but I can just catch a generic exception too

BradenBug: no, better to handle a more specific exception type for retries - might be worth specifically classifying which types of daytonaerrors are retryable: https://www.daytona.io/docs/python-sdk/common/errors#daytonaerror - what error modes are we addressing here?

BradenBug: for example, if we're running into rate limit errors, we would probably want an exponential backoff strategy

JarettForzano (approved):
LGTM

services/tracker/src/tracker/sandbox.py
```diff
@@ -353,6 +355,31 @@ async def archive_and_upload_output(
...
+            try:
+                await sandbox.refresh_data()
+                sandbox_state = sandbox.state
+            except Exception:
+                sandbox_state = "unknown (refresh failed)"
+
+            logger.warning(
+                f"exec failed (attempt {attempt + 1}/{max_attempts}): command={command}, "
+                f"sandbox={sandbox.name}, state={sandbox_state}, error={error}"
+            )
```
JarettForzano: Needs to be propagated to the log stream

tthuwng (commented):
Left a few implementation-scope comments on retry breadth, helper shape, and unrelated migration churn.

PR #239, opened 2026-04-01
feat: add org table and org-scoped data isolation

BradenBug:
## Summary

Adds organization-scoped data isolation as the foundation for multi-tenant support. All data queries are now filtered by org, with a "default" org for self-hosted backwards compatibility. This is Phase 1 of the Descope integration — auth will be added in #241.

## Changes
- Add `Org` model with `id` and `name` columns
- Add `org_id` FK to `benchmark`, `task`, `finalevaluation`, and `evaluationresult` tables
- Add Alembic migration that creates the org table with a dynamic UUID (not hardcoded), backfills existing rows, and sets NOT NULL constraints
- Add `scoped_select()`, `assert_org()`, and `get_scoped()` query helpers for consistent org filtering
- Add `get_current_org` FastAPI dependency (returns default org for now, Descope dispatch in #241)
- Thread org through all ~50 query sites in utils.py, all endpoints in main.py, and model methods
- Background task functions (`fetch_benchmark_row`, `fetch_task_row`) raise domain exceptions instead of HTTPException
- Default org resolved by name (`DEFAULT_ORG_NAME = "default"`) instead of hardcoded UUID
- Update all unit and integration test fixtures with org support

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor
- [ ] Docs / config

## Testing
- [x] Manually tested

## Checklist
- [x] Self-reviewed the diff
- [ ] Docs updated if needed


services/tracker/src/tracker/utils.py
```diff
@@ -549,6 +567,13 @@ async def process_benchmark(
             intervals=start_benchmark_request.webhook_intervals,
         )
 
+    # Resolve the org from the benchmark row (no org check on first fetch since the benchmark was just created by our system)
+    with Session(bind=engine) as session:
+        benchmark_row = session.get(Benchmark, benchmark_id)
+        if not benchmark_row:
+            raise TrackerServiceError(f"Benchmark with id {benchmark_id} not found")
+        org = session.exec(select(Org).where(Org.id == benchmark_row.org_id)).one()
+
```
JarettForzano, line 579: From the commit it looks like we are overriding this value right away on line 560-561. Since the `fetch_benchmark_row` method looks like it handles the check as well it looks like we can remove this

services/tracker/src/tracker/utils.py
```diff
@@ -838,8 +871,12 @@ async def stream_benchmark_results(
                     yield f"{EVENT_ERROR} {json.dumps({'error': 'Benchmark not found'})}\n\n"
                     break
 
+                if fresh_benchmark.org_id != org.id:
```
JarettForzano: You can add this statement to the conditional on line 870 `if not fresh_benchmark or fresh_benchmark.org_id != org.id: ...`

services/tracker/main.py
```diff
@@ -270,6 +274,9 @@ async def check_results_exist(
     Returns:
         {"exists": true/false}
     """
+    benchmark_row = session.get(Benchmark, benchmark_id)
+    assert_org(benchmark_row, org)
```
JarettForzano: Isn't this already supported by selecting the benchmark row with the given id and org? Instead of having a separate assert org method?

services/tracker/src/tracker/database/models.py
```diff
@@ -36,6 +36,14 @@
     )
 
 
+VALS_ORG_ID = UUID("00000000-0000-0000-0000-000000000001")
```
JarettForzano: Shouldn't this be done in the DB so that we are not giving people the id to our org? This would be sensitive info

services/tracker/src/tracker/database/scoping.py
JarettForzano, line 1: There seems to be a lot of cases of

```
benchmark_row = session.get(Benchmark, benchmark_id)
assert_org(benchmark_row, org)
```
happening, can you make a single method to consolidate the two?

services/tracker/src/tracker/database/models.py
JarettForzano: Since the org name is going to be unique we could just use that to fetch the org. Would this become obsolete in the next pr where we can create and fetch orgs based off descope?

BradenBug: Org is resolved from Descope API key - no endpoint accepts Org ID as input. User would need to authenticate with the Vals org key in order to access, so it's not really sensitive info. This wouldn't be obsolete in next PR because we still need a default org ID for self-hosted mode for backwards compatibility

services/tracker/src/tracker/database/scoping.py
BradenBug: Sure

services/tracker/src/tracker/utils.py
BradenBug: Fetch benchmark row requires org as a param, but we don't have org passed into this method, so we need to read the benchmark row first to determine which org it belongs to

services/tracker/src/tracker/utils.py
BradenBug: Sure

services/tracker/src/tracker/utils.py
JarettForzano: ah shoot yeah thats something i caused because i asked to separate the two prs

services/tracker/src/tracker/database/models.py
JarettForzano: Makes sense that we need it for backwards compatibility support, could we have the uuid assigned in the migration using uuid4()? I could have been more clear in the design doc so that is my bad. 

EDIT: Originally was thinking that we would also do a separate process like have the default org for anyone using the harness and then manually just change it to vals but that would probably be a lot of references to change so nvm.

JarettForzano: Lmk if you have any questions and after that we are good to merge this pr in, im going to hold off on doing another prod deploy till we have these two pr's in

JarettForzano (approved):
Approved but can you make sure that there is only one version head to the migrations/versions. I think from the back-merge another migration may have the same revision. Also turns out that the migrations don't get auto applied. You can fix it by using

`if echo "$@" | grep -q -e "tracker.serve" -e "uvicorn"; then` inside of the `entrypoint.sh` file

PR #251, opened 2026-04-07
Patch edge case with race condition

JarettForzano:
## Summary
There is an edge case which causes the benchmark to be stuck in a STOPPING status. This is caused when the worker exits before the force stop benchmark endpoint finishes running. The solution is simple, do not assume the worker task is running and instead clean up the final task manually inside of the endpoint.

## Changes
- At the end of the endpoint, check and update the final benchmark status if no tasks are being processed
- Add regression tests (confirmed applied patch works)

## Related Issues

Closes #

## Type of Change

- [ X ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [X  ] Refactor
- [ ] Docs / config

## Testing

- [ X] Added/updated unit tests
- [ X] Manually tested

## Checklist

- [ X] Self-reviewed the diff
- [ X] No debug/dead code left in
- [ ] Docs updated if needed

### DIFF
--- a/services/tracker/src/tracker/utils.py
+++ b/services/tracker/src/tracker/utils.py
@@ -973,6 +973,19 @@ async def force_stop_sandboxes(
         f"{task_alias}: {error_message}" for task_alias, error_message in results.items() if error_message
     )
 
+    # If all tasks are already in a stopped state, we need to update the final status here since the worker has exited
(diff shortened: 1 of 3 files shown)


services/tracker/src/tracker/utils.py
JarettForzano, line 1: Fix in this file

OrestesK (approved):

PR #255, opened 2026-04-07
Swap to signed url for install

### LINKED ISSUE #247: Use S3 signed url to download the agent directly into sandbox

JarettForzano:
## Summary
We were having large IO spikes when installing agents since each task needed to load the agent into memory and them unzip / copy it into the sandbox. Instead of doing that its much easier for us to just send the signed url into the sandbox and curl / unzip the agent. This way the agent never even passes through the tracker service.

This removes a source of resource spiking and allows us to lower the current resource cap in the future.

## Changes
- Install deps command that determines if curl or unzip need to be installed, if so goes through all of the package managers and finds the correct one to use
- curl and unzip inside of the same file instead of making multiple requests
- Create integration tests that confirm the command works on all of the base sandbox images that can be used

**Passed acceptance criteria**
Ran two complete benchmarks with no problems
- SWEbench
- TerminalBench
 
## Related Issues
Closes https://github.com/vals-ai/Valkyrie/issues/247

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ X] Refactor
- [ ] Docs / config

## Testing
- [X ] Added/updated unit tests
- [X ] Manually tested

## Checklist

- [ X] Self-reviewed the diff
- [X ] No debug/dead code left in
- [ ] Docs updated if needed

---

<a href="https://app.devin.ai/review/vals-ai/valkyrie/pull/255" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


services/tracker/src/tracker/sandbox.py
```diff
@@ -185,36 +182,58 @@ async def upload_agent_artifacts(
...
+    presigned_url = create_presigned_url(contract_s3_key, aws, s3_bucket)
+
+    zip_path = shlex.quote(f"/tmp/{contract.name}.zip")
+    contract_dir = shlex.quote(str(bundle_path / contract.name))
+    bundle_dir = shlex.quote(str(bundle_path))
+    quoted_url = shlex.quote(presigned_url)
+
+    # Install required dependencies inside of the instance
+    # Tracks if curl or unzip are missing and installs them if needed
+    install_deps = (
```
JarettForzano, line 210: I thought this would be weird to do but it works surprisingly well,
checks if the package manager is installed and if so runs the equivalent install command. Integration test confirms command is in the proper format

services/tracker/src/tracker/sandbox.py
```diff
@@ -185,36 +182,58 @@ async def upload_agent_artifacts(
...
+        f"curl -sfL -o {zip_path} {quoted_url}",
+        f"mkdir -p {bundle_dir}",
+        f"unzip -o -d {bundle_dir} {zip_path} -x contract.py",
+        f"rm -f {zip_path}",
+        f"mkdir -p {contract_dir}",
+    ]
+
+    script = " && ".join(steps)
+
+    result = await sandbox.process.exec(script)
```
BradenBug: nit: consider adding retries in case S3 download fails

BradenBug (approved):
looks good!

services/tracker/src/tracker/sandbox.py
JarettForzano: Yeah I can add a tenacity decorator

JarettForzano:
Will checkout why test is failing

BradenBug (approved):

PR #265, opened 2026-04-09
Add yaml config support

### LINKED ISSUE #246: Swap from python config to yaml

JarettForzano:
## Summary
Allow users to upload yaml configs as their contract instead of using python files. Its easier to use and can be validated upfront. This pr addresses the change while allowing backwards compatibility support. In the future pr ill be deprecating the python contract feature and adding a validate flag to the cli that allows users to validate their agent contract before even uploading it to aws.

## Error handling
image
image
image
image

## Tested agents

I converted two agents to yaml and uploaded them separately, use these when testing. You can also pull them down using `valk agent download <AGENT>` to see what the new contracts look like

- yaml_mini_sweagent
- yaml_codex

## Changes
- Moved method to parse python contract outside and instead made a switch statement between the python and yaml file parse methods
- Added a secondary path to create the agent contract request using the yaml config
- Added the agent contract as a base model under the schemas.py file, separating the different objects
- Validate run cmd on creation, enforces required values
- Added unit test that covers all possible edge cases I could think of
- Can also use choices with model to define a set of models that the agent allows

## Related Issues
Closes https://github.com/vals-ai/Valkyrie/issues/246

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactor
- [x] Docs / config

## Testing
- [x] Manually tested

## Checklist
- [x] Self-reviewed the diff
- [x] Docs updated if needed


src/valkyrie/schemas.py
JarettForzano, line 1: TODO: Errors need to be human readable, not the default pydantic ones

src/valkyrie/cli/bundler.py
```diff
@@ -95,36 +96,82 @@ def get_agent_zip_stream(agent_name: str | None, agent_path: Path) -> Generator[
...
     try:
         with tempfile.TemporaryDirectory() as tmp:
             tmp_path = Path(tmp)
-            contract_member = f"{agent_name}/contract.py"
             with zipfile.ZipFile(io.BytesIO(zip_bytes)) as zf:
-                zf.extract(contract_member, tmp_path)
-
-            return get_contract(tmp_path / contract_member, agent_config)
+                names = zf.namelist()
+                for ext in (".yaml", ".yml", ".py"):
```
JarettForzano, line 106: We have a few areas where we need to do this unfort, when we migrate completely this will be reverted

tests/unit/test_agent_contract.py
JarettForzano, line 1: Tests every combination of the yaml config, ill also be doing some more purposely try to break the config before I open the pr.

oalmatov (approved):

PR #285, opened 2026-04-16
Version control agents

### LINKED ISSUE #230: Create copy of agent inside of benchmark results folder we use after starting the benchmark, protects versioning during runtime
## Problem / Motivation
When you start a run and someone changes the agent it will break what is running for you. Solution is to version it by copying the agent into a separate folder that we reference while the run is ongoing

## Proposed Solution
Reuse the bucket folder for agent results and copy the agent itself inside when we start the run

JarettForzano:
## Summary
Instead of just using the agent that lives inside of `agents/<agent>.zip` we copy the agent into the `benchmarks/<id>` folder so that during the run any changes made to the agent are not affecting the in progress benchmarks. This also allows us to have a history if we ever need to go back and check what version of an agent was being ran with a benchmark.

## Changes

- Copy the agent into the benchmarks folder when the run starts
- Change all the paths to benchmarks/<benchmark-id>/<agent>.zip
- Update the tests
- Option to allow updating an agent version when resuming or retrying

```markdown
| `--update-agent, -u` | Refresh the frozen agent copy from the current `agents/<name>.zip` in S3 before resuming |
```

## Related Issues

Closes https://github.com/vals-ai/Valkyrie/issues/230

## Type of Change

- [x ] Bug fix
- [x ] New feature
- [ ] Breaking change
- [x ] Refactor
- [ ] Docs / config

## Testing

- [x ] Added/updated unit tests
- [x ] Manually tested

## Checklist

- [x ] Self-reviewed the diff
- [x ] No debug/dead code left in
- [ ] Docs updated if needed

---

<a href="https://app.devin.ai/review/vals-ai/valkyrie/pull/285" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


src/valkyrie/cli/main.py
```diff
@@ -768,6 +769,14 @@ def stop(run_id: UUID, force: bool):
     default=None,
     help="Path to a text file with one task ID per line",
 )
+@click.option(
```
BradenBug, line 772: I feel like we should just not have this. What use case are you imagining for it?

JarettForzano: Its important to have a way to update the version incase adhoc changes need to occur. For example if a rate limit error propagates and affects part of the run we would want to only rerun the tasks that it affected, not the whole thing. Without exposing a means to do this easily it will become a point of confusion-- such as, "The agent does not seem to be updated, I only want to rerun a few tasks" and so forth.

Up until now we have allowed changing the agent during a run so that functionality still needs to exist after the change, this is better since the user initiates the change instead of it just happening.

JarettForzano: TLDR; we don't want to remove functionality when we add features, since it can cause confusion and become a bottleneck later on. This pr preserves a way to update the agent after stopping the run, reserved for specific edge cases

BradenBug: Got it, didn't think about us being able to do this before

BradenBug (approved):

PR #292, opened 2026-04-17
feat: add distributed tracing to tracker and worker

BradenBug:
## Summary

Instruments the tracker API and Taskiq worker with OpenTelemetry, giving end-to-end visibility into benchmark execution from HTTP request through worker task completion. Trace context propagates across the Redis/Taskiq boundary so a single trace covers the full benchmark lifecycle. Traces and logs ship to Sentry.

Screenshot 2026-04-24 at 11 35 41 AM

## Architecture

- OTel SDK via the `logfire` package (spans + FastAPI/httpx auto-instrumentation). `send_to_logfire=False` — we don't ship to Logfire cloud, logfire is only used as an ergonomic OTel wrapper.
- **Traces** → Sentry via `SentrySpanProcessor`, with `SPAN_MAX_TIME_OPEN_MINUTES` bumped to 240 so long-running parents like `process_benchmark` / `process_task` survive (default 10m drops them).
- **Logs** → Sentry Logs via `LoggingIntegration(sentry_logs_level=INFO)`. Stdlib `logger.info/warning/error` appear alongside spans in the Sentry UI.
- **Propagation** via a composite propagator: W3C `traceparent`/`tracestate` + Baggage + Sentry's `sentry-trace`/`baggage`, so both non-Sentry peers (Daytona, benchmark_service) and Sentry peers see the trace.

## Changes

**Tracing config**
- New `tracing.py` with `configure_tracing(service_name)` — wires up the OTel SDK, Sentry span processor, composite propagator, and a `_ContextVarSpanProcessor` that attaches `request_id`/`benchmark_id`/`task_id` contextvars to every span (mirrors the existing `sentry._before_send` pattern for events).
- Auto-instrument FastAPI and httpx; SQLAlchemy and Redis intentionally excluded (high-volume, low-signal Taskiq plumbing).
- `/health` excluded from request tracing.

**Cross-boundary propagation**
- `/start-benchmark` and `/retry-or-resume-benchmark` `inject()` the current trace context into Taskiq labels before kicking.
- New `TracingContextMiddleware` (Taskiq) extracts trace context from labels on the worker side, so the kicked `process_benchmark` run shows up as a child of the originating HTTP request span.
- Sandbox-side: pass `DAYTONA_SANDBOX_OTEL_EXTRA_LABELS` env var so the sandbox's internal OTel telemetry is filterable by `benchmark_id`/`task_id`/`environment` (Daytona's OTLP export is account-level, so environment tags matter for separating dev/staging/prod).

**Manual spans**
- `process_benchmark` and `process_task` get top-level spans.
- Sub-op spans: `upload_agent`, `setup_task`, `run_agent`, `evaluate_instance`, `final_score`, `upload_results`, `create_log_group`, `upload_to_s3`, `send_notification`, `pty_disconnect`, `kill_pty_session`.
- Caught exceptions recorded via `logfire.exception()` at the three `sentry_sdk.capture_exception()` sites (`process_benchmark`, `process_task`, `TrackedTask.run`).

**Sentry config**
- `init_sentry` switched to `INSTRUMENTER.OTEL` and `enable_logs=True`.
- `LoggingIntegration(level=None, event_level=None, sentry_logs_level=INFO)` — breadcrumb capture and auto-error events disabled (spans carry context; we `capture_exception` explicitly).

**Infra**
- Removed `LOGFIRE_TOKEN` secret from `TrackerStack` / `WorkerStack` (no longer needed).
- docker-compose passes `SENTRY_DSN` / `SENTRY_RELEASE` / `ENVIRONMENT` through from host env for local testing.

## Follow-ups
- Instrument `benchmark_service` with OTel + Sentry if we want visibility into its internals — right now it appears only as outbound httpx spans from the tracker side, so you can see how long a call took but not what it was doing.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor
- [ ] Docs / config

## Testing
- [x] Manually tested

## Checklist
- [x] Self-reviewed the diff
- [ ] Docs updated if needed


services/tracker/src/tracker/config.py
```diff
@@ -54,21 +55,26 @@ def _build_database_url() -> str:
     redis_url=REDIS_URL,
 )
 
+# Order is load-bearing: TracingContextMiddleware must precede LoggingContextMiddleware
```
JarettForzano: Why is the order load-bearing?

services/tracker/docker-compose.yml
JarettForzano, line 1: Can remove the default for the region `AWS_REGION=${AWS_REGION:-us-east-1}`

JarettForzano (changes requested):
Instead of using context managers I think we should make decorators that wrap around the functions. Its easier to maintain and helps clean up the code by enforcing that we divide it up into functions.

services/tracker/src/tracker/config.py
BradenBug: Updated comment wording

services/tracker/src/tracker/utils.py
```diff
@@ -381,12 +394,23 @@ async def auto_flush_logs() -> None:
             task.status = TaskStatus.BUILDING
             task_session.commit()
 
+        environment = os.environ.get("ENVIRONMENT", "development")
```
JarettForzano: This I think should live inside of the config file

services/tracker/src/tracker/sentry.py
```diff
@@ -41,18 +33,28 @@ def init_sentry(service_name: str) -> None:
...
-            LoggingIntegration(
-                level=None,
-                event_level=None,
-            ),
-        ],
-    )
+    try:
+        sentry_sdk.init(
+            dsn=dsn,
+            environment=os.environ.get("ENVIRONMENT", "development"),
```
JarettForzano: Pull from config - just since its used in more than one location

services/tracker/src/tracker/logging/config.py
```diff
@@ -80,6 +80,9 @@ def configure_logging() -> None:
                     "filters": ["context"],
                 },
             },
+            # Sentry's LoggingIntegration (enabled in init_sentry with sentry_logs_level=INFO)
+            # attaches to the root logger and ships records to Sentry Logs without needing a
+            # dictConfig handler entry.
```
JarettForzano, line 85: it just uses the logs that we have built in?

services/tracker/src/tracker/sandbox.py
JarettForzano, line 1: I think we may want to use instrument with logfire on these so that we can capture the sandbox id. Should also use logfire.exception on the warnings so that we can make sure we are capturing those.

JarettForzano (approved):
Can merge after requested changes are made / responded to

services/tracker/src/tracker/sentry.py
BradenBug: sure can do

services/tracker/src/tracker/logging/config.py
BradenBug: Yep

services/tracker/src/tracker/sandbox.py
BradenBug: yuup

PR #296, opened 2026-04-20
Cap concurrent PTY WebSocket handshakes in tracker

oalmatov:
## Summary
Under load, Daytona begins failing PTY WebSocket handshakes above ~500 concurrent connects ([context](https://valsai.slack.com/archives/C0A4M9VB318/p1776699130377729?thread_ts=1776669359.717209&cid=C0A4M9VB318)). This PR gates PTY handshakes in the tracker service with a process-global semaphore capped at 100 — the ceiling that held 800 PTYs with zero failures in our testing. The semaphore only gates the WebSocket connect step; `handle.wait()`, `send_input`, and the rest of the PTY session lifetime remain ungated.

## Changes
- Add `_pty_handshake_slot` async context manager in `services/tracker/src/tracker/sandbox.py` backed by a module-level `asyncio.Semaphore(100)`. Scope is process-global, not per-benchmark, because the limit we care about is total in-flight handshakes hitting Daytona from the tracker process.
- Wrap `sandbox.process.create_pty_session` in `_create_pty_session` and `sandbox.process.connect_pty_session` in `_reconnect_and_wait_pty` with the new gate. Released immediately on return.
- Add observability logs: INFO when a task queues behind the cap (>100ms wait for a slot), WARNING when an individual handshake is slow (>2s). Both include current in-flight count out of cap (e.g. `inflight=97/100`) so CloudWatch filters can surface saturation directly.
- Add two unit tests (`services/tracker/tests/unit/test_sandbox.py`):
  - `test_semaphore_caps_concurrent_handshakes` — 25 concurrent calls against cap=5 never exceed the cap. Verified to fail if the gate is removed.
  - `test_semaphore_released_after_handshake_returns` — with cap=1, a task doing simulated post-handshake work must not block a second task's handshake. Verified to fail if the gate scope is widened beyond the handshake call.

## Related Issues

Closes #

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor
- [ ] Docs / config

## Testing
- [x] Manually tested

Unit tests cover cap enforcement and early release. Each test was also verified to fail under the specific regression it guards against (gate removed / gate scope widened).

Should deploy and do stress test on prod.

## Checklist
- [x] Self-reviewed the diff
- [ ] Docs updated if needed

### DIFF
--- a/services/tracker/src/tracker/sandbox.py
+++ b/services/tracker/src/tracker/sandbox.py
@@ -291,11 +291,9 @@ _PTY_CREATE_DELAY_SECONDS: float = 2.0
 # Process-global cap on concurrent PTY WebSocket handshakes.
 # Daytona starts failing above ~500 concurrent handshakes; 100 held 800 PTYs cleanly in testing.
 _PTY_HANDSHAKE_CAP: int = 100
-_PTY_HANDSHAKE_WAIT_LOG_THRESHOLD: float = 0.1
 _PTY_HANDSHAKE_SLOW_LOG_THRESHOLD: float = 2.0
 
 _pty_handshake_semaphore: Semaphore = Semaphore(_PTY_HANDSHAKE_CAP)
-_pty_handshake_inflight: int = 0
 
 # States that determine if the sandbox has been killed
 _DEAD_SANDBOX_STATES = (SandboxState.DESTROYING, SandboxState.DESTROYED, SandboxState.STOPPED)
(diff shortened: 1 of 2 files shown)


services/tracker/src/tracker/sandbox.py
```diff
@@ -287,10 +288,54 @@ async def install_agent_dependencies(
...
 _PTY_CREATE_DELAY_SECONDS: float = 2.0
 
+# Process-global cap on concurrent PTY WebSocket handshakes.
+# Daytona starts failing above ~500 concurrent handshakes; 100 held 800 PTYs cleanly in testing.
+_PTY_HANDSHAKE_CAP: int = 100
+_PTY_HANDSHAKE_WAIT_LOG_THRESHOLD: float = 0.1
+_PTY_HANDSHAKE_SLOW_LOG_THRESHOLD: float = 2.0
+
+_pty_handshake_semaphore: Semaphore = Semaphore(_PTY_HANDSHAKE_CAP)
+_pty_handshake_inflight: int = 0
```
OrestesK: pretty sure semaphore keep tracks of inflight alreayd, no need for new var

OrestesK (approved):

services/tracker/src/tracker/sandbox.py
oalmatov: just remove the counter entirely its not necessary. will just log when the sem is full. don't want to access a private attribute

PR #317, opened 2026-04-27
feat: forward descope key to benchmark services

BradenBug:
## Summary

Hosted benchmark services need to authenticate inbound calls from the tracker without reusing `X-Api-Key`, which is already reserved for Daytona sandbox credentials. This PR implements the tracker-side half of that flow: it copies the inbound Descope access key (`X-Api-Key`) onto outbound benchmark-service requests as `X-Descope-Api-Key`, so the receiving benchmark service can validate the user's Descope identity directly.

## Changes
- Add `forward_tracker_api_key()` in `services/tracker/src/tracker/auth.py`, which copies `service_headers` and injects the inbound tracker API key as `X-Descope-Api-Key`. Explicit downstream overrides are preserved (case-insensitive match), and a missing tracker key is a no-op.
- Wire the helper into `POST /start-benchmark` and `POST /retry-or-resume-benchmark/{benchmark_id}` so both initial runs and resumed runs carry the same downstream auth header.
- For resume, build the forwarded headers once and reuse them for both the immediate `BenchmarkServiceClient` and the persisted `start_benchmark_request` JSON, so background workers resuming the run also carry the key.
- Export `BENCHMARK_SERVICE_API_KEY_HEADER = "X-Descope-Api-Key"` as the single source of truth for the header name.

## Related
https://www.notion.so/vals-ai/Benchmark-Service-Descope-Authentication-34f8a877dd8d802fa41ac779222be710?t=new
https://github.com/vals-ai/create-benchmark-service/pull/30

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Refactor
- [ ] Docs / config

Not breaking: callers that already supply their own `X-Descope-Api-Key` in `service_headers` are preserved untouched, and callers that don't send `X-Api-Key` see no new headers added.

## Testing
- [x] Manually tested

New unit tests in `services/tracker/tests/unit/`:
- `test_auth.py`: `forward_tracker_api_key` adds the header, preserves an explicit override case-insensitively, and is a no-op when the tracker key is missing.
- `test_fastapi_server.py`: `/start-benchmark` forwards `X-Descope-Api-Key` into both the synchronous `BenchmarkServiceClient` headers and the queued `start_benchmark_request` JSON.
- `test_stop_and_resume.py`: `/retry-or-resume-benchmark/{id}` does the same on the resume path.

## Checklist
- [x] Self-reviewed the diff
- [ ] Docs updated if needed

## Follow-ups (not in this PR)
1. `create-benchmark-service` patch to validate `X-Descope-Api-Key` on HTTP and WebSocket routes.
2. Tenant allowlist enforcement (`DESCOPE_ALLOWED_TENANTS`).
3. Flip hosted benchmark-service deployments to `AUTH_REQUIRED=true` and remove static `BENCHMARK_API_KEY` after the Descope path is stable.

### DIFF
--- a/services/tracker/src/tracker/auth.py
+++ b/services/tracker/src/tracker/auth.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+from collections.abc import Mapping
+
 from descope import AuthException, DescopeClient
 from fastapi import Depends, HTTPException, Request
 from sqlmodel import Session, select
@@ -12,6 +14,8 @@ from tracker.database.session import get_session
 from tracker.logging import get_logger
 
 logger = get_logger(__name__)
+
+BENCHMARK_SERVICE_API_KEY_HEADER = "X-Descope-Api-Key"
 
 _cached_default_org: Org | None = None
 _descope_client: DescopeClient | None = (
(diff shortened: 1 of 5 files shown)


miyukimatthews (approved):

services/tracker/src/tracker/auth.py
```diff
@@ -62,6 +66,26 @@ def find_org_by_tenant(tenant_name: str, session: Session) -> Org | None:
...
+    """
+    forwarded_headers = dict(service_headers or {})
+    if not tracker_api_key:
+        return forwarded_headers
+
+    has_explicit_override = any(key.lower() == BENCHMARK_SERVICE_API_KEY_HEADER.lower() for key in forwarded_headers)
+    if not has_explicit_override:
+        forwarded_headers[BENCHMARK_SERVICE_API_KEY_HEADER] = tracker_api_key
+
+    return forwarded_headers
```
JarettForzano, line 86: I don't think we should be forwarding all of the headers, that would be a pretty big privacy concern

BradenBug: I agree we should not forward the full inbound request header set. This is just copying the existing explicit service_headers object and adding only the tracker auth credential as X-Descope-Api-Key.

PR #321, opened 2026-04-29
Support AWS session token in harness config

tthuwng:


tthuwng:
Closing: using a valid long-lived IAM user key in Valkyrie config instead of adding AWS_SESSION_TOKEN support.

PR #323, opened 2026-05-04
Clear stale final evaluation on retry

tthuwng:
## Summary
- clear stale `FinalEvaluation`, `finished_at`, and prior benchmark error state when retry/resume reopens tasks
- treat every non-terminal task status as blocking final benchmark finalization, including `EVALUATING`
- add focused tests for retry cleanup and final-status validation

## Root Cause
Retrying tasks inside an existing benchmark can set tasks back to `PENDING`/`IN_PROGRESS` while leaving the previous benchmark-level final evaluation visible. That lets `retrieve-results` expose stale final-score metadata while `fetch-benchmark` correctly reports the run as still in progress.

## Validation
- `uv run pytest tests/unit/test_stop_and_resume.py tests/unit/test_benchmark_utils.py -q`
- `uv run ruff check src/tracker/utils.py tests/unit/test_stop_and_resume.py tests/unit/test_benchmark_utils.py`
- `git diff --check`

### DIFF
--- a/services/tracker/src/tracker/utils.py
+++ b/services/tracker/src/tracker/utils.py
@@ -497,12 +497,14 @@ def set_benchmark_final_status(benchmark_row: Benchmark, session: Session, org: 
     Delegates status depending on if any tasks have been stopped.
     """
 
-    # Check if any tasks are still in the pending or in progress state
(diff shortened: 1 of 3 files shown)


tthuwng:
Closing per current VCB migration direction: no Valkyrie repo changes for this phase. We will keep the tracker/registry code path unchanged and focus only on minimal VCB benchmark service changes if needed.

PR #334, opened 2026-05-05
Add eval-only retry state

tthuwng:
## Summary

Adds tracker support for eval-only retry when a benchmark service has already saved evaluation progress state:

- stores latest `eval_resume_state` on `Task`
- allows `EVALUATING` tasks with saved state to resume through `BenchmarkServiceClient.resume_evaluation()` without creating a new sandbox
- falls back to normal scratch retry when no eval state exists
- records final eval results with nullable `EvaluationResult.instance_id`
- adds retry mode selection: default `auto`, plus `--from-scratch` to clear stored eval state and rerun from generation

## Dependency

Stacked on vals-ai/create-benchmark-service#38. The current dependency pin references create-benchmark-service commit `899b34aa39921e47e2a4cb5b1f0dbdb5ac89606b` from that PR.

### DIFF
--- a/services/tracker/src/tracker/database/models.py
+++ b/services/tracker/src/tracker/database/models.py
@@ -69,6 +69,11 @@ class AgentCausedExitReason(str, Enum):
     OS_KILLED = "OS_KILLED"
 
 
+class RetryMode(str, Enum):
+    AUTO = "auto"
+    FROM_SCRATCH = "from_scratch"
+
+
 class AgentContractRequest(BaseModel):
     name: str
     model: str | None = None
@@ -300,6 +305,7 @@ class Task(SQLModel, table=True):
     started_at: datetime = Field(default_factory=lambda: datetime.now(ZoneInfo("UTC")))
     error_message: str | None = Field(default=None)
     finished_at: datetime | None = None
+    eval_resume_state: dict[str, Any] | None = Field(default=None, sa_column=Column(JSON, nullable=True))
     benchmark: UUID = Field(foreign_key="benchmark.id")
 
     @computed_field
@@ -336,6 +342,6 @@ class EvaluationResult(SQLModel, table=True):
     id: UUID = Field(default_factory=uuid4, primary_key=True)
     org_id: UUID = Field(foreign_key="org.id")
     task: UUID = Field(foreign_key="task.id")
-    instance_id: str = Field(unique=True)
+    instance_id: str | None = Field(default=None, unique=True)
     agent_caused_exit_reason: AgentCausedExitReason | None = Field(default=None)
     result: dict[str, Any] = Field(default_factory=dict, sa_column=Column(JSON, nullable=False))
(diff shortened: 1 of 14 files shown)


services/tracker/src/tracker/utils.py
```diff
@@ -461,7 +504,11 @@ async def auto_flush_logs() -> None:
                 # NOTE: only really good for when we need to evaluate the container (for just evaluating a text response we can delegate before this)
                 logger.info(f"Evaluating agent {start_benchmark_request.contract.name} in sandbox {sandbox.name}")
                 evaluation_result = await benchmark_service.evaluate_instance(
-                    task_row.task_id, sandbox.id, on_message=log_output, dataset=start_benchmark_request.dataset
+                    task_row.task_id,
```
Chen-Oliver: still using evaluate_instance instead of evaluate_response?

services/tracker/src/tracker/utils.py
```diff
@@ -378,7 +386,42 @@ async def auto_flush_logs() -> None:
 
     flush_task = asyncio.create_task(auto_flush_logs())
 
+    def on_eval_resume_state(state: dict[str, Any]) -> None:
+        save_eval_resume_state(task_row.id, org, state)
+
     try:
+        if task_row.status == TaskStatus.EVALUATING:
+            assert task_row.eval_resume_state is not None
```
JarettForzano: Don't assert here, always want to raise exceptions

Chen-Oliver (approved):

JarettForzano (changes requested):
Update documentation

tthuwng:
Record from review:

One query -> Fetches for all eval resume states that exist and updates them to evaluating
Second query -> Fetches all tasks that are not evaluating and resets them

That should simplify this, it is also better to assume that if there is no resume state the user wants to retry it from scratch. We are currently just raising an error
https://github.com/vals-ai/Valkyrie/pull/334/changes#diff-f1b5ed8310e824bed2205dfe3662cfdf8fdf76355df2b863b6fe90f875b7d084R395

It won't be possible for people to know what tasks have a state stored inside or not

services/tracker/src/tracker/database/models.py
```diff
@@ -336,6 +342,6 @@ class EvaluationResult(SQLModel, table=True):
     id: UUID = Field(default_factory=uuid4, primary_key=True)
     org_id: UUID = Field(foreign_key="org.id")
     task: UUID = Field(foreign_key="task.id")
-    instance_id: str = Field(unique=True)
+    instance_id: str | None = Field(default=None, unique=True)
```
JarettForzano, line 345: In the future I want polymorphism to be used for things like this

JarettForzano (approved):

services/tracker/src/tracker/database/models.py
tthuwng: Agreed as follow-up tech debt. I kept this PR minimal with the existing retry-mode flag, but a separate polymorphic retry request object would be cleaner once we add more retry modes.

PR #358, opened 2026-05-08
feat: add run attribution to hosted mode

BradenBug:
## Summary

In hosted mode, capture the human (Descope access-key holder) who started each benchmark run and let users filter `valkyrie run list` by starter email. Self-hosted mode is unaffected — rows always have NULL `started_by_*`. Retry/resume preserves the original starter.

## Changes
- New nullable columns on `Benchmark`: `started_by_id` (Descope access key id) and `started_by_email` (lower-cased snapshot, indexed for filter).
- Alembic migration `d4e5f6a7b8c9` adds the columns + index.
- `auth.py`: new `RequestIdentity` value object, `resolve_descope_identity` helper, and `get_current_starter` FastAPI dependency. `get_current_org` becomes a one-line shim.
- `/start-benchmark` writes `started_by_*` from the JWT custom claims; `/init` returns `email_claim_missing` so the CLI can warn when a key isn't configured for attribution.
- `/fetch-benchmarks` accepts `started_by` as a list of emails (case-insensitive `IN` match); `BenchmarkTableRow` and `FetchBenchmarkMetadataResponse` expose `started_by_email`.
- CLI: `valkyrie run list --started-by alice@vals.ai,bob@vals.ai` (comma-separated, matches `--task-ids` convention) plus a "Started By" column in the output table. `valkyrie config init` prints a yellow warning when the access key has no `email` custom claim.
- Whitespace-only email claims are treated as missing (warns + writes NULL); filter inputs are stripped + lowercased.

## Related Issues
Closes #45

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor
- [ ] Docs / config

## Testing
- [x] Manually tested

## Checklist
- [x] Self-reviewed the diff
- [ ] Docs updated if needed

Screenshot 2026-05-14 at 8 18 03 PM


services/tracker/src/tracker/auth.py
```diff
@@ -58,7 +146,24 @@ def resolve_descope_tenant(api_key: str) -> str:
...
+        jwt_response.get(DESCOPE_ACCESS_KEY_ID_FIELD)
+    ) or _normalize_optional_string(
+        _get_descope_claim(jwt_response, "sub"),
+    )
+    if not access_key_id:
+        raise HTTPException(status_code=400, detail="Descope JWT missing access key id")
+
+    email = _normalize_optional_string(_get_descope_claim(jwt_response, "email"), lowercase=True)
+    name = _normalize_optional_string(_get_descope_claim(jwt_response, "name"))
+    user_id = _normalize_optional_string(_get_descope_custom_claim(jwt_response, DESCOPE_USER_ID_CLAIM))
```
JarettForzano: Call `_normalize_optional_string` inside of the method

services/tracker/src/tracker/auth.py
```diff
@@ -86,17 +191,33 @@ def forward_tracker_api_key(
...
-    Self-hosted (AUTH_REQUIRED=false): returns default org.
-    Hosted (AUTH_REQUIRED=true): validates Descope API key and resolves org.
+    Self-hosted (AUTH_REQUIRED=False): returns RequestIdentity with default org and Nones.
+    Hosted (AUTH_REQUIRED=True): validates Descope API key and resolves org + identity.
     """
+    if not AUTH_REQUIRED:
+        return RequestIdentity(org=get_default_org(session), access_key_id=None, email=None, name=None)
+
+    api_key = extract_api_key(request)
+    tenant_name, access_key_id, email, name = resolve_descope_identity(api_key, include_user_profile=True)
```
JarettForzano: What is access key here for?

services/tracker/src/tracker/auth.py
BradenBug: done

services/tracker/src/tracker/auth.py
BradenBug: it's persisted in DB as Benchmark.started_by_id to preserve the exact credential used to start the run

JarettForzano: This access key is just through descope right? Not similar to the AWS one? I have been trying to avoid saving any sort of credential in the DB if it can be avoided. If its descope I don't think its a big deal

BradenBug: yeah just descope id

services/tracker/src/tracker/auth.py
```diff
@@ -3,26 +3,116 @@
...
+        logger.warning("Failed to load Descope user profile for user_id=%s", user_id, exc_info=True)
+        return None, None
+
+    user = user_response.get("user")
+    if not isinstance(user, Mapping):
+        logger.warning("Descope user profile response did not include a user object for user_id=%s", user_id)
+        return None, None
+
+    email = _normalize_optional_string(user.get("email"), lowercase=True)
+    name = _normalize_optional_string(user.get("name")) or _normalize_optional_string(user.get("displayName"))
```
JarettForzano: `_normalize_optional_string(user.get("name") or user.get("displayName"))`

services/tracker/src/tracker/auth.py
```diff
@@ -3,26 +3,116 @@
...
+            continue
+
+        custom_claims = claim_source.get(DESCOPE_CUSTOM_CLAIMS_FIELD)
+        if isinstance(custom_claims, Mapping) and claim_name in custom_claims:
+            return _normalize_optional_string(custom_claims.get(claim_name), lowercase=lowercase)
+
+    return None
+
+
+def _load_descope_user_profile(user_id: str) -> tuple[str | None, str | None]:
```
JarettForzano: No need for a tuple if you just are taking whatever is available after

```
if include_user_profile and email is None and user_id is not None:
        email, profile_name = _load_descope_user_profile(user_id)
        name = name or profile_name
```

services/tracker/src/tracker/auth.py
```diff
@@ -43,8 +133,15 @@ def extract_api_key(request: Request) -> str:
     return api_key
 
 
-def resolve_descope_tenant(api_key: str) -> str:
-    """Validate an API key against Descope and return the tenant name (no DB lookup)."""
+def resolve_descope_identity(
+    api_key: str, *, include_user_profile: bool = False
+) -> tuple[str, str, str | None, str | None]:
```
JarettForzano: You can return an object here instead of a tuple if 4 is too many

services/tracker/main.py
```diff
@@ -168,15 +170,16 @@ def init_org(
...
+    return {"org_name": org.name, "created": created, "email_claim_missing": email is None}
 
 
 @app.post("/start-benchmark")
 async def start_benchmark(
     http_request: Request,
     request: StartBenchmarkRequest,
     session: Session = Depends(get_session),
-    org: Org = Depends(get_current_org),
+    starter: RequestIdentity = Depends(get_current_starter),
```
JarettForzano: Is starter an abbreviation of something?

services/tracker/main.py
```diff
@@ -204,6 +207,12 @@ async def start_benchmark(
     )
     logger.info(f"Starting benchmark run - contract: {request.contract.name}, benchmark: {request.benchmark_name}")
 
+    if starter.access_key_id is not None and starter.email is None:
+        logger.warning(
+            "Access key %s has no 'email' custom claim; run attribution for this run will be empty",
+            starter.access_key_id,
+        )
+
```
JarettForzano: Should we log this after the benchmark row is fetched so that we can get the benchmark id too or will the middleware take care of that for us?

services/tracker/src/tracker/auth.py
JarettForzano, line 1: Can we have doc strings under the methods in here? And a example of what a custom claim from descope may look like. It helps visualize what we are parsing

JarettForzano:
Do we need to update documentation to include the new key?

services/tracker/main.py
BradenBug: it's basically the whole identity (org + user) of who started the run

services/tracker/main.py
BradenBug: Ur right, I'll move it after

services/tracker/src/tracker/auth.py
BradenBug: done

JarettForzano (approved):
Approved but flush changes from dev to prod before merging this to dev

PR #359, opened 2026-05-09
infra: bump worker min/max tasks to 2/4

Chen-Oliver:
## Summary
- Set `WORKER_MIN_TASKS=2` so a single restart/deploy never leaves zero healthy workers (yesterday's 5/7 incident hit a 20-min blackout when the fleet briefly dropped to 1 and that one task died).
- Bump `WORKER_MAX_TASKS=2 → 4` so the existing CPU-based autoscaler has room to react. Note: the deployed `desiredCount` has been running at 3-6 out of band, so the application-autoscaling target on `service/AgenticHarnessCluster/Worker` should be reconciled before deploy (otherwise this would force a scale-in).

## Caveat / follow-up
This is a bandaid. CPU is a poor scaling signal for the worker because it's mostly I/O-bound (await-ing Daytona sandbox execution and LLM calls), and queue depth is the wrong granularity (one taskiq job = one benchmark run, which fans out internally to many concurrent tasks). The proper fix is a custom metric — e.g. total tasks-in-flight from the tracker DB, or a worker-emitted active-task gauge — fed into a `scale_on_metric` target-tracking policy. Tracking this separately; the constants change here just buys availability headroom in the meantime.

## Test plan
- [ ] Verify current application-autoscaling min/max on the Worker service via `aws application-autoscaling describe-scalable-targets --service-namespace ecs --resource-ids service/AgenticHarnessCluster/Worker` before deploying.
- [ ] `cdk diff` the worker stack and confirm only the autoscaling target min/max change.
- [ ] After deploy, confirm `RunningTaskCount` on the Worker service stays at 2 during quiet periods and can scale up to 4 under load.


infra/constants.py
```diff
@@ -42,8 +42,8 @@
 # Worker Service
 WORKER_CPU = 4096
 WORKER_MEMORY = 8192
-WORKER_MIN_TASKS = 1
-WORKER_MAX_TASKS = 2
+WORKER_MIN_TASKS = 2
+WORKER_MAX_TASKS = 4
```
BradenBug, line 46: number go up

BradenBug (approved):

PR #361, opened 2026-05-12
feat: add DEFERRED task status for staged runs

Chen-Oliver:
## Summary

Adds a new `TaskStatus.DEFERRED` so a single `benchmark_id` can register the full dataset up front but only execute a subset of tasks. The remaining tasks sit in `DEFERRED` until explicitly promoted — avoiding the "subset run, then full run from scratch" pattern that re-runs tasks already completed.

## Usage

```bash
# Initial cheap subset run; the rest are registered as DEFERRED
valkyrie run start --benchmark X --task-ids \"t1,t2,t3\" --defer-rest

# Default retry/resume does NOT touch DEFERRED (same protection as ERROR)
valkyrie run resume <id> --retry

# Promote specific deferred tasks individually
valkyrie run resume <id> --task-ids \"t4\"

# Or promote all remaining deferred tasks at once
valkyrie run resume <id> --run-deferred
```

## Design

- DEFERRED is excluded from the worker's runnable filter and from the default \`retry_statuses\`, so it stays inert unless the caller explicitly names tasks (\`--task-ids\`) or passes \`--run-deferred\`.
- A benchmark with DEFERRED tasks still transitions to \`FINISHED\` once every non-deferred task completes — DEFERRED behaves analogously to ERROR at the benchmark-status level. **No \`BenchmarkStatus\` changes required.**
- Migration adds the enum value to Postgres via \`ALTER TYPE ... ADD VALUE\`.

## Why a new status (vs reusing STOPPED)

We considered repurposing STOPPED for parked tasks, but STOPPED is in the default \`retry_statuses\` list — so any retry-or-resume call would accidentally sweep up parked tasks. A dedicated DEFERRED status keeps the opt-in semantics clean.

## Test plan

Tested end-to-end against a real benchmark service + agent + LLM (\`openai/gpt-5.4-nano-2026-03-17\` on a 3-task dev set):

- [x] \`run start --task-ids X --defer-rest\` → 1 PENDING + 2 DEFERRED rows; benchmark transitions to FINISHED with deferred tasks present
- [x] \`run resume --retry\` → ERROR task picked up, DEFERRED tasks untouched
- [x] \`run resume --task-ids Y\` → only Y promoted from DEFERRED → PENDING
- [x] \`run resume --run-deferred\` → all remaining DEFERRED tasks promoted
- [x] Normal path (no \`--defer-rest\`, no \`--task-ids\`) → unchanged: all dataset tasks PENDING, zero DEFERRED rows

---

<a href="https://app.devin.ai/review/vals-ai/valkyrie/pull/361" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/services/tracker/src/tracker/utils.py
+++ b/services/tracker/src/tracker/utils.py
@@ -931,6 +931,9 @@ class BenchmarkContext:
     def _task_counts(self) -> TaskCounts:
         finished_statuses = [TaskStatus.FINISHED, TaskStatus.ERROR, TaskStatus.STOPPED]
 
+        # Exclude DEFERRED from both totals — those tasks were intentionally not queued
+        # and shouldn't drag down progress (e.g. a subset-only run with 180/180 done
+        # should show 100%, not 180/450). They'll re-enter the counts when promoted.
         statement = (
             select(
                 func.count().label("total_tasks"),
@@ -940,6 +943,7 @@ class BenchmarkContext:
             .select_from(Task)
             .where(Task.benchmark == self._benchmark_row.id)
             .where(Task.org_id == self._org.id)
+            .where(Task.status != TaskStatus.DEFERRED)
         )
 
         result = self._session.exec(statement).one()
@@ -1025,8 +1029,9 @@ def catch_errors_during_cleanup(benchmark_id: UUID, session: Session, org: Org) 
     if benchmark_row.status in terminal_statuses:
         return
 
-    # Force non exited tasks to be ERROR
-    task_terminal_statuses = [TaskStatus.FINISHED, TaskStatus.ERROR, TaskStatus.STOPPED]
+    # Force non exited tasks to be ERROR. DEFERRED tasks were intentionally
+    # not queued, so leave them alone — they should still be promotable later.
+    task_terminal_statuses = [TaskStatus.FINISHED, TaskStatus.ERROR, TaskStatus.STOPPED, TaskStatus.DEFERRED]
     session.exec(
         update(Task)
         .where(col(Task.benchmark) == benchmark_id)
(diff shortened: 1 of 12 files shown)


services/tracker/src/tracker/utils.py
```diff
@@ -1199,8 +1224,15 @@ async def force_stop_sandboxes(
         f"{task_alias}: {error_message}" for task_alias, error_message in results.items() if error_message
     )
 
-    # If all tasks are already in a stopped state, we need to update the final status here since the worker has exited
-    finished_statuses: list[TaskStatus] = [TaskStatus.FINISHED, TaskStatus.ERROR, TaskStatus.STOPPED]
+    # If all tasks are already in a stopped state, we need to update the final status here since the worker has exited.
+    # DEFERRED counts as terminal here too — those tasks were never queued, so they don't block the benchmark from
+    # transitioning out of STOPPING.
```
JarettForzano, line 1229: ✂️

services/tracker/src/tracker/utils.py
```diff
@@ -911,6 +931,9 @@ def _status(self) -> BenchmarkStatus:
     def _task_counts(self) -> TaskCounts:
         finished_statuses = [TaskStatus.FINISHED, TaskStatus.ERROR, TaskStatus.STOPPED]
 
+        # Exclude DEFERRED from both totals — those tasks were intentionally not queued
+        # and shouldn't drag down progress (e.g. a subset-only run with 180/180 done
+        # should show 100%, not 180/450). They'll re-enter the counts when promoted.
```
JarettForzano, line 936: ✂️

JarettForzano (commented):
- Is it necessary to pre-compute the entire dataset instead of allowing the user to adjust it themselves?
- Does this support use cases where the size of a dataset is increased and we want to run the remaining tasks and calculate the new final score?
- What is the benefit to having the deferred status, the only time that we use it is on retry? We could just request the remaining task ids then without a lot of the additional functionality added in this pr.

```python
verify_response = await benchmark_service.verify_task_ids(
    task_ids=request.task_ids, slice_str=request.slice_str, dataset=request.dataset
)

# If --defer-rest, fetch the full dataset task list so the benchmark also stages
# the unselected tasks as DEFERRED rows. They sit dormant until promoted via
# retry-or-resume (or named explicitly in --task-ids).
if request.defer_rest:
    full_response = await benchmark_service.verify_task_ids(
        task_ids=None, slice_str=None, dataset=request.dataset
    )
    all_dataset_task_ids = full_response.task_ids
```

src/valkyrie/cli/utils.py
```diff
@@ -161,6 +162,7 @@ def format_task_breakdown(task_breakdown: dict[TaskStatus, int]) -> str:
             TaskStatus.EVALUATING,
             TaskStatus.ERROR,
             TaskStatus.STOPPED,
+            TaskStatus.DEFERRED,
```
JarettForzano, line 165: Should go at the top, before pending (this is the order we use that is displayed in the cli btw)

PR #363, opened 2026-05-12
feat(tracker): lazily add tasks on resume via --task-ids

Chen-Oliver:
## Summary
- `retry-or-resume` now creates fresh `PENDING` rows for any `--task-ids` that don't already have a row, as long as they're valid in the current dataset (validated via the benchmark service's existing `verify_task_ids` call).
- Lets you add work to a `STOPPED` or `FINISHED` run without starting over — including tasks added to the dataset after the run began.
- Replaces the tracker's "missing task_id" guard with the benchmark service's existing verify response (invalid ids surface as 400 from the service, propagated as 500 by the tracker).

This supersedes #361 (which introduced an explicit `DEFERRED` status + `--defer-rest` / `--run-deferred` flags) — lazy creation covers the same UX with one mechanism instead of three.

## Notes for reviewers
- Close #361 — that branch is superseded by this approach (no DEFERRED status, no extra flags, smaller code surface).

---

<a href="https://app.devin.ai/review/vals-ai/valkyrie/pull/363" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/services/tracker/src/tracker/utils.py
+++ b/services/tracker/src/tracker/utils.py
@@ -666,7 +666,7 @@ def create_task_rows(
     task_ids_to_create = [task_id for task_id in verified_task_ids if task_id not in existing_task_ids]
 
     for task_id in task_ids_to_create:
-        task_row = Task(org_id=org.id, task_id=task_id, benchmark=benchmark_row.id, status=TaskStatus.PENDING)
+        task_row = Task(org_id=org.id, task_id=task_id, benchmark=benchmark_row.id)
         session.add(task_row)
 
     session.commit()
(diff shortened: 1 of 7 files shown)


services/tracker/src/tracker/utils.py
```diff
@@ -664,7 +666,7 @@ def create_task_rows(
     task_ids_to_create = [task_id for task_id in verified_task_ids if task_id not in existing_task_ids]
 
     for task_id in task_ids_to_create:
-        task_row = Task(org_id=org.id, task_id=task_id, benchmark=benchmark_row.id)
+        task_row = Task(org_id=org.id, task_id=task_id, benchmark=benchmark_row.id, status=TaskStatus.PENDING)
```
JarettForzano: default when created should be pending

BradenBug (approved):

services/tracker/src/tracker/utils.py
Chen-Oliver:
```suggestion
        task_row = Task(org_id=org.id, task_id=task_id, benchmark=benchmark_row.id)
```

PR #366, opened 2026-05-13
feat(cli,tracker): http URL task-ids files + subset rescore on results

Chen-Oliver:
## Motivation

Teams often maintain a **curated subset** of a benchmark — a smaller, representative slice of task IDs that they care about reporting on (e.g., a published headline number) — separately from the full task suite the agent is actually run against. The canonical subset definition typically lives in a shared, version-pinnable place: a `.txt` of task IDs in a GitHub repo, an S3 object, a published URL.

Today, scoring an agent on both the full suite and the curated subset from the *same* run isn't supported:
- The tracker computes one `FinalEvaluation` over whatever tasks the run included, and the S3 final view reflects that.
- To get a subset score you'd either start a second run scoped to the subset (duplicating compute) or compute the subset score yourself off-line (no benchmark-service scoring guarantees, no consistent artifact).

This PR closes that gap with the smallest possible surface change: a subset filter on `run results`, plus URL support for the `--task-ids-file` flag so the subset definition can live anywhere reachable.

## What this enables

```bash
# Subset definition lives in a versioned remote location:
#   https://raw.githubusercontent.com/<org>/<repo>/<sha>/subsets/<bench>.txt

# Start a run on the full benchmark (your existing flow, unchanged)
valk run start --agent ... --benchmark <bench> --dataset <full-set>

# Once it finishes, get the full-set score and the curated-subset score from the same run
valk run results <id>                                                            # full
valk run results <id> --task-ids-file https://.../subsets/<bench>.txt            # subset
```

Pairs naturally with the lazy-spike pattern landed in #363: start small with the curated subset, extend the same run to the full suite by passing the full task list to `run resume`, and afterwards query both scores from a single run.

## Changes

- `--task-ids-file` on `run start` / `run resume` / `run results` accepts an `http(s)` URL in addition to a local path (uses `urllib.request` — no new deps). Lets users share a curated task-id list without local setup.
- `run results --task-ids` / `--task-ids-file` filters the FinalView to that subset and re-calls `benchmark_service.final_score` over the filtered `evaluation_results`; the persisted `FinalEvaluation` and per-task rows are left untouched (DB is source of truth).
- CLI prints `Scored over <K> of <N> subset task ids` so stale / partially-overlapping subsets surface visibly (e.g. a versioned subset file that names tasks added after the run started).

## Subset-score semantics

The recompute calls `benchmark_service.final_score` with only the filtered `evaluation_results`. Whether the resulting score is subset-relative or dataset-relative depends on the benchmark service's `calculate_final_score` implementation (some normalize against the full dataset, others against what was passed). Benchmark-service maintainers can update on their side if subset-relative semantics matter for their score.

## Test plan

- [x] 150 tracker unit tests pass; new test (case 9 in `test_retrieve_results`) exercises the subset path with a mocked `final_score` and asserts the `X-Descope-Api-Key` forwarding header
- [x] 40 CLI unit tests pass
- [x] ruff format + check clean
- [x] live end-to-end against a local tracker + benchmark service + python http server
  - `run start --task-ids-file <URL>` → run created with tasks fetched from URL
  - `run resume --task-ids-file <URL>` → lazy spike adds new PENDING rows; all complete
  - `run results` (no flags) → full set scored
  - `run results --task-ids a,b` → "Scored over 2 of 2"
  - `run results --task-ids-file <URL>` → "Scored over N of N"
  - `run results --task-ids a,missing-id` → "Scored over 1 of 2" warning fires
  - `run results --task-ids X --task-ids-file Y` → `--task-ids and --task-ids-file are mutually exclusive`

### DIFF
--- a/src/valkyrie/cli/utils.py
+++ b/src/valkyrie/cli/utils.py
@@ -5,10 +5,13 @@ import json
 import shutil
 import tarfile
 import tempfile
+import urllib.request
+from collections.abc import Iterable
 from datetime import datetime
 from enum import Enum
 from pathlib import Path
 from typing import Any, Coroutine, TypeVar
+from urllib.parse import urlparse
 from uuid import UUID
 
 import click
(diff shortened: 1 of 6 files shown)


src/valkyrie/cli/utils.py
JarettForzano, line 1: Is the point of this http / https thing for S3? This kinda feels like a bit overkill

BradenBug, line 1: out of scope for this change, but I think we should start breaking up these "utils" files into better scoped modules with less responsibilities

BradenBug: The http/https part only adds like 3 lines of code, the rest is a refactor out of main.py, I wouldn't call it overkill

BradenBug (approved):
run integ tests

src/valkyrie/cli/utils.py
Chen-Oliver: not related to S3.
this is just so we can put all the task id lists (swebench index, fab index, vcb index) on github and pull from there instead of needing people to have those files locally

PR #374, opened 2026-05-14
Allow active retry of errored tasks

tthuwng:
## Summary
- allow the existing retry/resume endpoint to accept `retry=true` while a benchmark is still `IN_PROGRESS`
- when a benchmark is running, only retry tasks currently in `ERROR`; running resume without retry is a no-op
- restrict resumed worker task rows to the exact verified task IDs so active retry cannot pick unrelated runnable work
- reload task state from the DB before final scoring so active retries are not scored as stale failures

## Ticket
Addresses vals-ai/valkyrie-ticket-library#65.


JarettForzano (changes requested):
- Can we add a new endpoint for this? I think the current one is already too feature packed and we might need to revisit it down the line. 
- Are we still able to guarantee changes are atomic with this functionality?

tthuwng:
> * Can we add a new endpoint for this? I think the current one is already too feature packed and we might need to revisit it down the line.
> * Are we still able to guarantee changes are atomic with this functionality?

I think keeping it on the same endpoint is more convenient for callers because the intent is still “retry this benchmark.” The route can stay stable and Valkyrie can choose the right behavior from benchmark state:

- stopped/error run: existing retry/resume behavior
- in-progress run + retry=true: retry only tasks currently in ERROR
- in-progress run + no errored tasks: no-op
- stopping run: reject

So callers do not need to learn a second endpoint or decide which route matches the benchmark lifecycle state. The important part is keeping the implementation narrow: active retry should not add another general-purpose mode to the endpoint. It should be one explicit branch with a tight invariant: while running, only ERROR tasks can be reset and the spawned worker only receives those exact task IDs.

I agree the endpoint is feature-packed and worth revisiting later, but I’d treat that as an API cleanup/refactor, not a reason to split this behavior now.


On atomicity: within the tracker DB, yes for the task reset. We verify the selected tasks first, then update only eligible ERROR tasks and clear their old eval rows in one commit before enqueueing the worker. The worker is scoped to the exact verified task IDs.

services/tracker/src/tracker/utils.py
```diff
@@ -824,6 +827,12 @@ async def process_benchmark(
             evaluation_results=evaluation_results, dataset=start_benchmark_request.dataset
         )
 
+        with Session(bind=engine) as session:
+            benchmark_row = fetch_benchmark_row(benchmark_id, session, org)
+            if has_runnable_tasks(session, benchmark_row, org):
+                finalization_deferred = True
+                return
```
JarettForzano, line 862: This looks like a duplicate check, what was the reasoning for adding this in?

services/tracker/src/tracker/utils.py
```diff
@@ -1241,38 +1251,59 @@ async def reset_to_in_progress_status(
...
+                retry_statuses.append(TaskStatus.ERROR)
+
+            filter_query = [
+                col(Task.benchmark) == benchmark_row.id,
+                col(Task.org_id) == org.id,
+                or_(
+                    col(Task.status).in_(retry_statuses),
+                    col(Task.task_id).in_(rerun_task_ids),
+                ),
+            ]
```
JarettForzano: Can you put this into a separate method that just returns the filter query? That way its not all inlined and under a if else statement. Instead of using an if else statement we should use a fallback since there is only one case.

> if status == progress : return x
> 
> return y

services/tracker/src/tracker/utils.py
```diff
@@ -1241,38 +1251,59 @@ async def reset_to_in_progress_status(
...
+                    existing_by_task_id[task_id] for task_id in rerun_task_ids if task_id in existing_by_task_id
+                ]
+                missing_task_ids = [task_id for task_id in rerun_task_ids if task_id not in existing_by_task_id]
+                if missing_task_ids:
+                    raise TrackerServiceError(
+                        f"{', '.join(missing_task_ids)} cannot be retried while run {benchmark_row.id} is in progress because they are not in ERROR status"
+                    )
+            new_task_ids = []
+        else:
+            new_task_ids = [tid for tid in rerun_task_ids if tid not in existing_by_task_id]
```
JarettForzano, line 1338: Somethings up with the way this is structured, can you take another look at it?

services/tracker/src/tracker/utils.py
```diff
@@ -1241,38 +1251,59 @@ async def reset_to_in_progress_status(
...
+        all_requested_task_ids = [task.task_id for task in existing_rows] + new_task_ids
         verify_response = await benchmark_service.verify_task_ids(
             task_ids=all_requested_task_ids, slice_str=None, dataset=benchmark_row.arguments.dataset
         )
 
-        # Set the benchmark status to in progress to flag resuming the benchmark
-        benchmark_row.status = BenchmarkStatus.IN_PROGRESS
-        session.add(benchmark_row)
-        session.commit()
+        if benchmark_row.status != BenchmarkStatus.IN_PROGRESS:
```
JarettForzano, line 1352: "Can already be in progress if we retry tasks while the run is ongoing" - Leave comment for edge case

services/tracker/src/tracker/utils.py
```diff
@@ -1241,38 +1251,59 @@ async def reset_to_in_progress_status(
...
+            new_task_ids = [tid for tid in rerun_task_ids if tid not in existing_by_task_id]
 
         # Allow re-running the end of the benchmark without running any tasks
         if not existing_rows and not new_task_ids:
             return []
 
         # Verify the task ids are still valid before priming to resume
         # Raises if any task ids are invalid
-        all_requested_task_ids = list(existing_by_task_id.keys()) + new_task_ids
+        all_requested_task_ids = [task.task_id for task in existing_rows] + new_task_ids
```
JarettForzano, line 1346: If this is done to preserve the order, doesn't .keys() already do that?

services/tracker/main.py
```diff
@@ -512,26 +512,23 @@ async def retry_or_resume_benchmark(
...
-
-    if benchmark_row.status in invalid_states:
+    if benchmark_row.status == BenchmarkStatus.STOPPING:
         raise HTTPException(
             status_code=400,
-            detail=f"Run {benchmark_id} is in the {benchmark_row.status} state. Cannot continue a run that is currently running.",
+            detail=f"Run {benchmark_id} is in the {benchmark_row.status} state. Cannot continue a run that is stopping.",
+        )
+
+    benchmark_is_running = benchmark_row.status == BenchmarkStatus.IN_PROGRESS
```
JarettForzano: Inline that

services/tracker/main.py
```diff
@@ -544,6 +541,17 @@ async def retry_or_resume_benchmark(
...
+        return RetryOrResumeBenchmarkResponse(
+            status="success",
+        )
+
+    # NOTE: 0 is not acceptable
+    if concurrency:
+        benchmark_row.arguments.concurrency = concurrency
+        session.add(benchmark_row)
+        session.commit()
+
```
JarettForzano: Validation needs to happen before we mutate the state of any object instance. What was the reasoning for moving this?

services/tracker/src/tracker/utils.py
```diff
@@ -672,34 +669,50 @@ def create_task_rows(
...
+        if status == TaskStatus.FINISHED:
+            assert result is not None
+            evaluation_results[task_id] = result
+            continue
+        if status in [TaskStatus.ERROR, TaskStatus.STOPPED]:
+            evaluation_results[task_id] = None
+            continue
+        raise TrackerServiceError(f"Unexpected task status {status} while finalizing run {benchmark_row.id}")
+
+    return evaluation_results
```
JarettForzano: Can be simplified

services/tracker/src/tracker/utils.py
```diff
@@ -69,6 +69,7 @@
 
 _SANDBOX_CREATION_CAP: int = 10
 _PTY_TASK_RETRY_LIMIT: int = 1
+RUNNABLE_TASK_STATUSES = [TaskStatus.PENDING, TaskStatus.BUILDING, TaskStatus.IN_PROGRESS, TaskStatus.EVALUATING]
```
JarettForzano: Private this

JarettForzano (changes requested):
I think there are TOCTOU problems with this but it might not be a blocking issue. The biggest problem is that we are optionally ending the process benchmark task. Due to the way that the queue system is designed for the tasks to run we don't really have a choice.

Right now the tasks are all being threaded inside of the same process but after this change, I might need to change it to use taskq so that we can use IPC and track the tasks across multiple processes. If you ran this 3-5 times, we want to guarantee it works 100% of the time. After you clean this up a bit we can get it in. Have you tested this out to confirm base case works well?

services/tracker/src/tracker/utils.py
tthuwng: it's to catches a retry that starts while `final_score` is in flight, before writing FinalEvaluation

services/tracker/src/tracker/utils.py
tthuwng: for stopped/error resume, implicit stopped/error rows stay first, and requested rows are appended in requested order. So I now build all_requested_task_ids from the intentionally ordered existing_rows

services/tracker/src/tracker/utils.py
```diff
@@ -702,34 +699,50 @@ def create_task_rows(
...
+        match status:
+            case TaskStatus.FINISHED:
+                assert result is not None
+                evaluation_results[task_id] = result
+            case TaskStatus.ERROR | TaskStatus.STOPPED:
+                evaluation_results[task_id] = None
+            case _:
+                raise TrackerServiceError(f"Unexpected task status {status} while finalizing run {benchmark_row.id}")
+
+    return evaluation_results
```
JarettForzano: Do you think this is redundant? Using an outer join returns None if the evaluation result does not exist. The `has_runnable_tasks` method above `fetch_final_score_inputs` should validate if all the tasks are in an expected state.

services/tracker/src/tracker/utils.py
```diff
@@ -1369,6 +1389,25 @@ async def reset_to_in_progress_status(
         raise TrackerServiceError(f"Unexpected error resuming run {benchmark_row.id}: {str(e)}") from e
 
 
+def _retry_task_filters(benchmark_row: Benchmark, retry: bool, rerun_task_ids: list[str], org: Org) -> list[Any]:
```
JarettForzano, line 1390: This is a good improvement, can you leave a doc string?

The main restriction that your feature has is that you cannot retry tasks that have been finished. So if you see an error but the task was marked as finished, you would need to wait till the run is finished to retry that task.

services/tracker/src/tracker/utils.py
```diff
@@ -1303,38 +1314,47 @@ async def reset_to_in_progress_status(
...
+                existing_rows = [existing_by_task_id[task_id] for task_id in rerun_task_ids]
+            new_task_ids = []
+        else:
+            if rerun_task_ids:
+                implicit_rows = [task for task in existing_rows if task.task_id not in rerun_task_ids]
+                requested_rows = [
+                    existing_by_task_id[task_id] for task_id in rerun_task_ids if task_id in existing_by_task_id
+                ]
+                existing_rows = implicit_rows + requested_rows
+            new_task_ids = [tid for tid in rerun_task_ids if tid not in existing_by_task_id]
```
JarettForzano, line 1338: I took a look into this, I think I can confirm that this logic overlaps with the `_retry_task_filters` filter. The only thing you have that is needed would be this.

```python
if benchmark_row.status == BenchmarkStatus.IN_PROGRESS:
    missing = [tid for tid in rerun_task_ids if tid not in existing_by_task_id]
    if missing:
        raise TrackerServiceError(
            f"{', '.join(missing)} cannot be retried while run {benchmark_row.id} is in progress because they are not in ERROR status"
        )
```

services/tracker/main.py
```diff
@@ -517,26 +517,25 @@ async def retry_or_resume_benchmark(
...
+        )
+
+    if benchmark_row.status == BenchmarkStatus.IN_PROGRESS and not retry:
+        return RetryOrResumeBenchmarkResponse(
+            status="success",
         )
 
+    if concurrency is not None and concurrency < 1:
+        raise HTTPException(status_code=400, detail="Concurrency must be greater than 0.")
+
```
JarettForzano, line 612: Lol might have been better before, since we just reuse the previous concurrency if they don't specify it inside of the request. Fair to remove unless I am missing something

JarettForzano: I hid the comment on the concurrency since I don't think its applicable anymore

JarettForzano (changes requested):
Final round of review, will approve after all comments are addressed / responded to

services/tracker/src/tracker/utils.py
tthuwng: Agreed. Simplified `fetch_final_score_inputs` to rely on the outer join plus the preceding `has_runnable_tasks` guard. It now selects only `task_id` and `EvaluationResult.result`, casts the query shape, and returns `dict(task_rows)`.

services/tracker/src/tracker/utils.py
tthuwng: Added the docstring. It now explicitly says active retries on in-progress runs are limited to `ERROR` tasks, and finished tasks need to wait until the run is terminal.

services/tracker/src/tracker/utils.py
tthuwng: Agreed. Removed the overlapping row reshuffle and left the in-progress branch with just the missing-task validation. `_retry_task_filters` now owns selecting eligible rows.

services/tracker/main.py
tthuwng: Ack, leaving concurrency handling unchanged since the hidden comment is no longer applicable.

JarettForzano (approved):
Approved, thanks for addressing

PR #408, opened 2026-05-20
[codex] Fix resume concurrency persistence

tthuwng:
## Summary
- persist resume concurrency overrides by replacing the JSON-backed benchmark arguments object
- add regression coverage that the queued resume request and stored benchmark args both use the override

## Why
ProgramBench stop/resume with --concurrency 20 initially accepted the request, but the tracker row kept the old concurrency because the nested JSON field was mutated in place.

## Validation
- uv run pytest services/tracker/tests/unit/test_stop_and_resume.py -q

### DIFF
--- a/services/tracker/main.py
+++ b/services/tracker/main.py
@@ -639,7 +639,7 @@ async def retry_or_resume_benchmark(
         )
 
     if concurrency is not None:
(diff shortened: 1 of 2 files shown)


tthuwng:
Closing per request.

PR #432, opened 2026-06-01
Update CBS lock and provider lifecycle handling

tthuwng:
## Summary
- Update root and tracker uv.lock to create-benchmark-service 0.7.1 at merged CBS main commit c82ef50.
- Pulls in vals-ai/create-benchmark-service#58, which restores Daytona delete lifecycle conflict retry/normalization in the provider.
- Translate provider sandbox errors from sandbox.command back into tracker SandboxError so crash/lifecycle handling remains provider-generic.
- Use a 100 item sandbox list page size for force-stop, matching the old Daytona force-stop behavior.
- Keep force-stop cleanup open for a few short passes so Daytona list/delete and in-flight sandbox creates settle before the caller continues.
- Clean up sandbox creates that finish after task cancellation by waiting for the provider create call to resolve, deleting that sandbox, and then preserving cancellation.

## Validation
- uv run pytest services/tracker/tests/unit/test_sandbox.py services/tracker/tests/unit/test_stop_and_resume.py -q, 39 passed.
- uv run ruff check services/tracker/src/tracker/sandbox.py services/tracker/src/tracker/utils.py services/tracker/tests/unit/test_sandbox.py services/tracker/tests/unit/test_stop_and_resume.py.
- uv run ruff format --check services/tracker/src/tracker/sandbox.py services/tracker/src/tracker/utils.py services/tracker/tests/unit/test_sandbox.py services/tracker/tests/unit/test_stop_and_resume.py.
- Cross-package regression through tracker.utils.stop_sandbox with locked CBS main passed: delete conflict is retried and stop_sandbox returns None.
- Tried removing the short force-stop settle loop entirely; live tracker integration failed in test_force_stop_sandboxes because callers continued while sandbox contexts were still observing deleted containers. Kept the loop but simplified it to stop on the first empty list pass.
- GitHub tracker integration workflow manually rerun on attempt 2 after the expected PR guard: passed, 27 integration tests in 3m1s.
- All PR checks are green on head c8c4a04.

### DIFF
--- a/services/tracker/src/tracker/sandbox.py
+++ b/services/tracker/src/tracker/sandbox.py
@@ -1,5 +1,6 @@
 """Sandbox management utilities for the tracker service."""
 
+import asyncio
 import base64
 import shlex
 import time
@@ -7,7 +8,7 @@ import uuid
 from asyncio import Semaphore
 from collections import deque
 from collections.abc import Callable
-from contextlib import asynccontextmanager
+from contextlib import asynccontextmanager, suppress
 from pathlib import PurePosixPath
 from typing import Any, AsyncGenerator
 
@@ -182,7 +183,16 @@ async def create_sandbox(
     try:
         async with creation_semaphore:
             start = time.monotonic()
-            sandbox = await _create_sandbox(provider, sandbox_name, source, resources, labels, env_vars)
+            create_task = asyncio.create_task(
+                _create_sandbox(provider, sandbox_name, source, resources, labels, env_vars)
+            )
+            try:
+                sandbox = await asyncio.shield(create_task)
+            except asyncio.CancelledError:
+                with suppress(Exception):
+                    sandbox = await create_task
+                    await delete_sandbox(sandbox, provider)
+                raise
     except Exception as e:
         incr("valkyrie.sandbox.create.errors", tags={"error_class": type(e).__name__})
         raise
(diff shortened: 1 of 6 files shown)


services/tracker/src/tracker/sandbox.py
```diff
@@ -182,7 +183,16 @@ async def create_sandbox(
...
+            create_task = asyncio.create_task(
+                _create_sandbox(provider, sandbox_name, source, resources, labels, env_vars)
+            )
+            try:
+                sandbox = await asyncio.shield(create_task)
+            except asyncio.CancelledError:
+                with suppress(Exception):
+                    sandbox = await create_task
+                    await delete_sandbox(sandbox, provider)
+                raise
```
JarettForzano, line 195: What is the reasoning behind making a coroutine around the `create_sandbox` method?

services/tracker/src/tracker/utils.py
```diff
@@ -1315,12 +1320,26 @@ async def force_stop_sandboxes(
...
+                results[sandbox.name] = result
+
+            if found_sandbox:
+                empty_passes = 0
+            else:
+                empty_passes += 1
+                if empty_passes >= 2:
+                    break
+
+            await asyncio.sleep(_SANDBOX_FORCE_STOP_SETTLE_SECONDS)
```
JarettForzano: Can we simplify this?

services/tracker/src/tracker/sandbox.py
tthuwng: This is specifically to handle force-stop cancellation while the provider create call is already in flight. A direct await can be cancelled before tracker receives the sandbox handle, even though Daytona may still finish creating the sandbox server-side. By keeping the create as a task and shielding it, we can wait for the in-flight create to resolve on cancellation, delete the sandbox if it was created, then re-raise the cancellation.

services/tracker/src/tracker/utils.py
tthuwng: Yep, simplified this. I removed the extra empty_passes state and now the loop is just: list/delete sandboxes for a bounded number of passes, break on the first empty pass, and only sleep after a pass that actually found sandboxes. I reran the focused local tests and manually reran tracker integration after the PR guard; it passed 27 integration tests in 3m1s on c8c4a04.

PR #458, opened 2026-06-10
fix(tracker): grade workspace on non-zero agent exit instead of erroring

dnialh:
## Problem
`run_agent` (`stream_command_output`, `services/tracker/src/tracker/sandbox.py`) branches on the agent's exit code:
- `SUCCESS` → continue to evaluation
- `TIMEOUT` (124) / `OS_KILLED` (137) → return an `AgentCausedExitReason` → **continue to evaluation**
- **any other non-zero → `raise AgentRunFailedError` → task errored, evaluation skipped**

So when the agent **self-crashes** — most commonly `MaxContextWindowExceededError` (exit 1) on large-codebase tasks — the task is discarded as a hard error, even though the agent left a perfectly gradeable partial migration in `/submission`. The `AgentCausedExitReason` docstring literally says these are "exit reasons caused by the agent that continue to evaluation" — a context-window crash is exactly that, it just wasn't whitelisted.

In the recent code-migration campaign this silently threw away **~42** real (low-scoring) attempts as "errors" instead of grading them.

## Fix
Treat a generic non-zero agent exit the same as timeout/OOM:
- Add `AgentCausedExitReason.AGENT_ERROR`.
- In `stream_command_output`, **return** `AGENT_ERROR` (logging the exit code + last output, keeping the sentry tag) instead of raising.
- `run_agent` logs a warning for `AGENT_ERROR` and proceeds to evaluation, mirroring TIMEOUT/OS_KILLED.

Genuine infra failures are unaffected — `ProviderSandboxError` / `SandboxNotFoundError` are caught and raised **above** this branch; reaching the exit-code branch means the command actually ran and returned a code (the gradeable case).

## Tests
Updated the exit-code test to assert the new return-and-continue behavior (returns `AGENT_ERROR`, tags `agent_exit_code`, surfaces last output via `on_output`). `ruff` + `basedpyright` clean; **167 tracker unit tests pass**.

### DIFF
--- a/services/tracker/src/tracker/database/models.py
+++ b/services/tracker/src/tracker/database/models.py
@@ -75,6 +75,9 @@ class AgentCausedExitReason(str, Enum):
 
     TIMEOUT = "TIMEOUT"
     OS_KILLED = "OS_KILLED"
+    # Agent self-terminated non-zero (e.g. crashed on a context-window limit).
+    # Agent-caused, not infra — the workspace is still gradeable, so continue.
+    AGENT_ERROR = "AGENT_ERROR"
 
 
 class RetryMode(str, Enum):
--- a/services/tracker/src/tracker/sandbox.py
+++ b/services/tracker/src/tracker/sandbox.py
@@ -44,7 +44,6 @@ from tracker.database.models import (
     OutputArtifactSpec,
 )
 from tracker.exceptions import (
-    AgentRunFailedError,
     OutputArtifactError,
     SandboxError,
     SandboxSetupError,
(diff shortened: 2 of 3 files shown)


dnialh:
Superseded by the agent-side fix: `mini_swe_code_migration` origin/main (PR #3, commit 83b9721) already rewrites the run_cmd to catch model-limit crashes (MaxContextWindowExceededError|LimitsExceeded|TimeExceeded|model query timed out) and force status=0 so the captured workspace grades. The tracker can't be modified in this environment; the agent fix is the correct, deployable layer. Closing (reopen if a tracker-side change is ever wanted).

PR #463, opened 2026-06-14
Remove Python contract support (YAML-only contracts)

oalmatov:
## Summary

All agent contracts are now authored as `contract.yaml`, so this PR removes the legacy Python contract (`contract.py` / `BaseAgentContract`) code path entirely — YAML becomes the only supported format. Every agent in the registry has already been migrated, smoke-tested, and re-pushed (see vals-ai/agent-registry#35), so nothing depends on the Python path. Keeping it meant maintaining two parsers, dynamically `exec`-ing arbitrary contract code at load time (`importlib`), and `.py`/`.yaml`/`.yml` branching throughout the CLI. YAML-only is simpler and safer.

This is the **base** of a 3-PR stack (→ #465 → #466).

## Changes

- **Delete** `src/valkyrie/contract.py` (the `BaseAgentContract` ABC).
- `cli/bundler.py`: remove `_parse_python_contract` and the `.py` branches in `get_contract` / `get_contract_from_zip_bytes`; drop the `importlib.util` / `BaseAgentContract` imports. `get_contract` accepts only `.yaml`/`.yml`.
- `cli/s3_client.py`: drop the `importlib`-based `.py` fallback in `get_ingest_lambda_from_s3`; `ingest_lambda` is read directly from `contract.yaml`.
- `cli/main.py`: `valk run start` on a local dir resolves `contract.yaml`/`.yml` only; updated the `analyze` error message.
- `tracker/sandbox.py`: drop the now-dead `unzip -x contract.py` exclusion (and the matching assertion in the artifact-upload integration test).
- Docs/comments updated to YAML-only (`docs/CONTRACTS.md`, `tracker/types.py`).
- Second commit removes pre-existing dead code paths in the bundler surfaced during review.
- Full sweep confirms no remaining references to `BaseAgentContract`, `valkyrie.contract`, `_parse_python_contract`, dynamic contract import, or `contract.py`.

## Related Issues

- Agent migration: vals-ai/agent-registry#35
- Part of the stack that supersedes the original combined PR (old #430).

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Breaking change
- [x] Refactor
- [x] Docs / config

## Testing

- [x] Added/updated unit tests
- [x] Manually tested

All registry agents were converted to YAML and smoke-tested (slice `0:1`) on their representative benchmarks before this removal.

## Checklist

- [x] Self-reviewed the diff
- [x] No debug/dead code left in
- [x] Docs updated if needed

PR #430, opened 2026-06-01
Valkyrie: tracker api for platform integration

oalmatov:
## Summary
Tracker REST API + DB schema for the platform-managed Valkyrie integration. Tracker is stateless with respect to identity and org config: it reads AWS creds / S3 bucket / Daytona pointer from `X-Harness-*` request headers and run attribution from `X-Harness-Run-By-Email`, populated by the platform frontend (see vals-ai/platform#2406).

## Changes
- New REST endpoints under `/agents`, `/benchmark-services`, `/benchmarks`, `/filter-options`, `/benchmarks/status`, `/benchmarks/{id}/tasks`.
- Dropped tracker-owned `User` and `OrgConfig` tables; added `Benchmark.run_by_email` + composite index for keyset pagination.
- Dual-mode Descope auth: `Authorization: Bearer` for web sessions, `x-api-key` for the CLI.
- Benchmark-service auth secret resolution via `service_auth_header_name` / `service_auth_secret_name` on `StartBenchmarkRequest`.
- Pre-flight checks (`health_check` + `verify_task_ids`) run before DB row creation so failed-auth / unreachable runs do not pollute the list.
- CLI: agent bundler + server-side contract resolution.
- Review follow-ups: simplified `/benchmarks/status` empty handling, expanded `benchmark_id` naming, simplified CSV token parsing, formatter cleanup after presigned URL helper renaming, restored the auth settings marker, switched bearer parsing to FastAPI’s auth-header helper, and restored the benchmark status TODO comment.

## Related Issues
N/A

## Type of Change
- [x] Bug fix
- [x] New feature
- [x] Breaking change
- [x] Refactor
- [ ] Docs / config

## Testing
- [x] Added/updated unit tests
- [x] Manually tested

Local validation:
- `make format-check`
- `make lint`
- `make typecheck`
- `cd services/tracker && uv run pytest tests/unit/test_auth.py tests/unit/test_api_review_endpoints.py` — 15 passed

CI: all normal checks pass, including Codecov patch, ruff, basedpyright, tracker unit tests, CLI tests, and smoke test. `tracker-integration-tests` exits early on PRs with `Check must be run manually on PRs`.

## Checklist
- [x] Self-reviewed the diff
- [x] No debug/dead code left in
- [x] Docs updated if needed

Link to Devin session: https://app.devin.ai/sessions/00c95b8e959c421fa66d6dbc53d1db9d
Requested by: @oalmatov

PR #465, opened 2026-06-15
Valkyrie: switch tracker S3 to aioboto3 + CLI consolidation

oalmatov:
## Summary

Refactors the tracker's S3 layer onto **aioboto3** (genuinely async, instead of sync boto3 wrapped in `asyncio.to_thread`) and consolidates the CLI's duplicated S3 logic onto the tracker's helpers. This is "ground-prep" — it changes no product behavior, but gives the platform-integration API (#466) a clean async S3 layer to build on.

Stacked on **#463**.

## Changes

**Commit 1 — `tracker: switch S3 layer to aioboto3`**
- `download_from_s3`, `delete_from_s3`, `s3_object_exists`, `list_s3_objects`, `create_presigned_url` are now `async` (native aioboto3); `_s3_client` opens an aioboto3 client (session cached per credential set).
- Callers in `main.py` / `sandbox.py` updated to `await`; the agent-output tar endpoint is now an async generator (per-file `download_from_s3` + `BytesIO`, dropping `download_from_s3_stream`).
- `handle_s3_error` is **async-aware** (it previously did nothing on coroutine functions — the error only surfaced after the wrapper returned) and fully typed with `ParamSpec`, so `s3.py` typechecks clean (15 → 0 errors).
- Adds `aioboto3` to the tracker's dependencies.

**Commit 2 — `valkyrie cli: consolidate S3 key + copy logic onto tracker helpers`**
- Replace hardcoded S3 key strings with `get_contract_s3_key` / `get_benchmark_contract_s3_key` / `S3_BENCHMARKS_PREFIX`.
- Route `update_benchmark_agent_version` through the tracker's (now-async) `copy_s3_object`, with an existence pre-check to preserve the friendly "agent not found" message.
- Single source of truth for the bucket layout.

## Related Issues

- Base: #463 · Next: #466
- Part of the stack superseding the original combined PR (old #430).

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor
- [ ] Docs / config

## Testing

- [x] Manually tested

Tracker unit suite (185) and CLI unit suite (51) pass; `basedpyright` clean on `s3.py` and no new errors in the converted callers. Test mocks for the converted functions updated to async.

## Checklist

- [x] Self-reviewed the diff
- [x] Docs updated if needed


oalmatov:

- #463
    - #465 ◀
        - #466

services/tracker/src/tracker/aws/s3.py
```diff
@@ -1,37 +1,45 @@
...
         aws_secret_access_key=aws.aws_secret_access_key,
         aws_session_token=aws.aws_session_token,
         region_name=aws.aws_default_region,
-        config=Config(max_pool_connections=200),
     )
 
 
+def _s3_client(aws: "AWSCredentials") -> Any:
+    """Open an async S3 client for the given credentials (use as `async with`)."""
+    return _s3_session(aws).client("s3", config=_CLIENT_CONFIG)  # pyright: ignore[reportUnknownMemberType]
```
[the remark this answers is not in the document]
oalmatov: Fixed. Added `download_many_from_s3` (one shared client/connection pool reused across the loop) and switched the tar paths to it, restoring the connection reuse the cached sync `boto3` client had before the aioboto3 switch. A process-global cached client was avoided on purpose — it would break the per-test event loops. (Note: the loop downloads one tar per task, so the reuse matters most for full-benchmark fetches with many tasks.)

services/tracker/main.py
```diff
@@ -759,31 +760,33 @@ async def fetch_agent_outputs(
...
                 try:
-                    body, size = download_from_s3_stream(s3_key, harness_config.aws, harness_config.s3_bucket)
+                    data = await download_from_s3(s3_key, harness_config.aws, harness_config.s3_bucket)
 
                     tarinfo = tarfile.TarInfo(name=relative_path)
-                    tarinfo.size = size
+                    tarinfo.size = len(data)
 
-                    tar.addfile(tarinfo, fileobj=body)
+                    tar.addfile(tarinfo, fileobj=io.BytesIO(data))
```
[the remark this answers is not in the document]
oalmatov: Addressed via the same `download_many_from_s3` helper: it reads one object into memory at a time, so peak memory is bounded by the largest single object rather than the whole set (documented in the helper's docstring).

services/tracker/src/tracker/aws/s3.py
```diff
@@ -102,14 +121,34 @@ def download_from_s3(s3_key: str, aws: "AWSCredentials", s3_bucket: str) -> byte
...
+async def download_many_from_s3(
+    s3_keys: Iterable[str], aws: "AWSCredentials", s3_bucket: str
+) -> AsyncIterator[tuple[str, bytes]]:
+    """Download multiple objects over a single shared client (one connection pool).
+
+    Best-effort: keys that fail to download are logged and skipped. Yields
+    ``(s3_key, content)`` for each successful download. Each object is read fully
+    into memory one at a time, so peak memory is bounded by the largest single
+    object rather than the whole set.
+    """
```
oalmatov, line 138: used to download one output per task, now does it in a batch. one output per task may be over defensive, lmk what u think

services/tracker/src/tracker/aws/s3.py
```diff
@@ -47,10 +58,24 @@ def get_agent_result_s3_key(benchmark_id: str, task_id: str, output_name: str) -
     return f"{S3_BENCHMARKS_PREFIX}/{benchmark_id}/{task_id}/{output_name}"
 
 
-def handle_s3_error(message: str):
-    def decorator(func):
+def handle_s3_error(message: str) -> Callable[[Callable[_P, _R]], Callable[_P, _R]]:
+    """Wrap AWS errors as S3Error. Works on both sync and async callables."""
```
oalmatov: same decorator but just async aware

JarettForzano: You can remove sync support and remove the decorator from the two that use it (they don't make api calls)

```python
@handle_s3_error(message="Failed to create run URL")
def create_benchmark_url(benchmark_id: str, region: str, s3_bucket: str) -> str: ...
```

```python
@handle_s3_error(message="Failed to create console URL")
def create_console_url(s3_key: str, region: str, s3_bucket: str) -> str: ...
```

services/tracker/src/tracker/aws/s3.py
JarettForzano: good

services/tracker/src/tracker/aws/s3.py
```diff
@@ -225,21 +238,19 @@ def list_s3_objects(prefix: str, aws: "AWSCredentials", s3_bucket: str) -> list[
...
-    paginator = client.get_paginator("list_objects_v2")
-
     object_keys: list[str] = []
-    for page in paginator.paginate(Bucket=s3_bucket, Prefix=prefix):
-        if "Contents" in page:
-            for obj in page["Contents"]:
-                if "Key" in obj:
-                    object_keys.extend([obj["Key"]])
+    async with _s3_client(aws) as client:
+        paginator = client.get_paginator("list_objects_v2")
```
JarettForzano: I suggest turning this into an async iterator so that we are not loading everything into memory. you can yield the keys (ik it was not like this before, but figured that we can do this during the refactor)

JarettForzano:
Looks great, only small stuff

services/tracker/src/tracker/aws/s3.py
oalmatov: Done — removed `@handle_s3_error` from `create_console_url`/`create_benchmark_url` (pure string builders, no API calls) and made the decorator async-only (dropped the sync branch + the `inspect`/`cast` imports). All remaining users are async.

services/tracker/src/tracker/aws/s3.py
oalmatov: Done — `list_s3_objects` is now an async iterator that `yield`s keys (no full list in memory). `download_many_from_s3` accepts a sync/async iterable so it streams lazily from it, and `fetch_agent_outputs` peeks a single key (to keep the empty→404 before the stream starts) then streams the rest through one shared client.

PR #466, opened 2026-06-15
Valkyrie: platform-integration REST API (schema · auth · api)

oalmatov:
## Summary

Adds the FastAPI surface the platform UI consumes from the tracker — the actual platform-integration feature. Built on the async (aioboto3) S3 layer from #465. Organized as three reviewable commits: schema → auth → API.

Stacked on **#465** (top of the stack).

## Changes

**Commit 1 — schema**
- Add `Benchmark.run_by_email` + a composite `ix_benchmark_org_started_at_id` index, via a single **add-only** Alembic migration (chains off the current dev head; no spurious `user`/`org_config` drops).

**Commit 2 — auth**
- Dual-mode Descope auth: `Authorization: Bearer` for web sessions and `x-api-key` for the CLI, resolving to the same org.

**Commit 3 — API + wiring**
- New endpoints: `/agents`, single-benchmark, single-task, `/benchmark-services`, `/benchmarks/status`, `/filter-options`.
- CORS middleware, keyset pagination, server-side contract resolution from S3, and the agent bundler moved into the tracker (`tracker.agent_bundler`; the CLI bundler is now a thin shim).
- `list_agents` consolidated into one shared async helper used by both the `/agents` endpoint and the `valk agent list` CLI command.
- New endpoints are written against the async S3 layer (`await`).
- Unit + integration test suites for the new endpoints.

## Related Issues

- Base: #465 (← #463)
- Supersedes the original combined PR (old #430), split into this stack.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor
- [ ] Docs / config

## Testing

- [x] Manually tested

Tracker unit suite (185) and CLI unit suite (51) pass; integration suites cover the new endpoints. `basedpyright` shows no new errors beyond the pre-existing sqlmodel-typing items already present on the prior branch.

## Checklist

- [x] Self-reviewed the diff
- [x] Docs updated if needed


oalmatov:

- #463
    - #465
        - #466 ◀

services/tracker/main.py
```diff
@@ -657,7 +759,7 @@ async def retry_or_resume_benchmark(
         session.commit()
 
     if concurrency is not None:
-        benchmark_row.arguments = benchmark_row.arguments.model_copy(update={"concurrency": concurrency})
+        benchmark_row.arguments.concurrency = concurrency
```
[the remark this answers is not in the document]
oalmatov: Good catch — fixed. Now reassigns: `benchmark_row.arguments = benchmark_row.arguments.model_copy(update={'concurrency': concurrency})`, matching the adjacent resume-secrets pattern, so SQLAlchemy detects the change on the JSON-backed TypeDecorator column.

services/tracker/src/tracker/api/benchmarks_status.py
```diff
@@ -0,0 +1,61 @@
...
+            BenchmarkStatusEntry(
+                id=benchmark.id,
+                status=benchmark.status,
+                finished_at=benchmark.finished_at,
+                total_tasks=sum(counts.values()),
+                finished_tasks=(
+                    counts.get(TaskStatus.FINISHED, 0)
+                    + counts.get(TaskStatus.ERROR, 0)
+                    + counts.get(TaskStatus.STOPPED, 0)
+                ),
```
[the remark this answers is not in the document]
oalmatov: Aligned, but upward rather than down: STOPPED is treated as terminal everywhere else in the codebase (utils.py `finished_statuses`/`terminal_statuses` = FINISHED+ERROR+STOPPED). So the inconsistent ones were single_benchmark.py and build_benchmark_table_rows, which omitted STOPPED — I added it there. All three endpoints now report finished_tasks = FINISHED+ERROR+STOPPED, so a stopped run reads as terminal instead of stuck.

services/tracker/src/tracker/auth.py
```diff
@@ -259,19 +316,3 @@ def get_current_starter(request: Request, session: Session = Depends(get_session
             detail=f"Organization '{identity.tenant_name}' not configured — run valk config init",
         )
     return RequestIdentity(org=org, access_key_id=identity.access_key_id, email=identity.email, name=identity.name)
```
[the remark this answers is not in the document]
oalmatov: Intentional, as you surmised. `get_current_starter` needs the Descope `exchange_access_key(include_user_profile=True)` flow for run attribution (email/name), which the Bearer session path doesn't carry. The web FE passes `x-api-key` for /start-benchmark specifically.

services/tracker/pyproject.toml
```diff
@@ -51,7 +51,7 @@ build-backend = "hatchling.build"
 prerelease = "allow"
 
 [tool.uv.sources]
-create-benchmark-service = { git = "https://github.com/vals-ai/create-benchmark-service.git", rev = "v0.7.4" }
+create-benchmark-service = { git = "https://github.com/vals-ai/create-benchmark-service.git", rev = "v0.7.2" }
```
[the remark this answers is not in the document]
oalmatov: Fixed — bumped back to v0.7.4. This was an accidental downgrade: the feature branch was based on an older dev that still pinned v0.7.2, and the merge carried it over v0.7.4. Restored to match the base branch + regenerated the lock.

services/tracker/src/tracker/api/benchmark_services.py
```diff
@@ -0,0 +1,51 @@
...
+    start = time.monotonic()
+    try:
+        resp = await client.get(health_url)
+        latency_ms = int((time.monotonic() - start) * 1000)
+        return BenchmarkServiceHealth(
+            name=name,
+            url=url,
+            healthy=resp.status_code == 200,
+            latency_ms=latency_ms,
+            error=None if resp.status_code == 200 else f"HTTP {resp.status_code}",
```
JarettForzano, line 32: This seems a bit weird, if you just return the status_code you know theres an error if status_code != 200

services/tracker/src/tracker/api/filter_options.py
```diff
@@ -0,0 +1,38 @@
...
+
+from fastapi import APIRouter, Depends
+from pydantic import BaseModel
+from sqlmodel import Session, select
+
+from tracker.auth import get_current_org
+from tracker.database.models import Benchmark, Org
+from tracker.database.session import get_session
+
+router = APIRouter(prefix="/benchmarks")
```
JarettForzano, line 13: I suggest organizing apis based off folders, like api/benchmarks/ that way you can separate different endpoints into different modules but keep them under the same subpackage. That should be easier to parse than having a single api folder with all the endpoints inside

services/tracker/src/tracker/api/single_benchmark.py
```diff
@@ -0,0 +1,108 @@
...
+
+    statuses = parse_csv(status, TaskStatus)
+    base_filters = [
+        col(Task.benchmark) == benchmark_id,
+        col(Task.org_id) == org.id,
+    ]
+    if len(statuses) == 1:
+        base_filters.append(col(Task.status) == statuses[0])
+    elif len(statuses) > 1:
+        base_filters.append(col(Task.status).in_(statuses))
```
JarettForzano: Can't you use _in for both?

services/tracker/src/tracker/aws/s3.py
```diff
@@ -307,3 +308,37 @@ def create_benchmark_url(benchmark_id: str, region: str, s3_bucket: str) -> str:
...
     return f"https://{region}.console.aws.amazon.com/s3/buckets/{s3_bucket}?region={region}&prefix={prefix}"
+
+
+@handle_s3_error(message="Failed to list agents from S3")
+async def list_agents(aws: "AWSCredentials", s3_bucket: str) -> list[tuple[str, datetime | None]]:
+    """List agents under the `agents/` prefix.
+
+    Accepts both layouts:
+      * subfolders — `agents/<name>/...`
+      * zipped bundles — `agents/<name>.zip`
```
JarettForzano: If the agents are currently stored as .zip, why do we support subfolders?

.gitignore
```diff
@@ -31,3 +31,6 @@ build
 
 *claude*
 *CLAUDE*
+
+# Local docker-compose overrides (per-dev port remaps, etc.)
+services/tracker/docker-compose.override.yml
```
JarettForzano, line 36: Local testing?

tests/unit/test_tracker_service.py
JarettForzano, line 1: Looks like the override secrets unit test got removed

JarettForzano (changes requested):
Great work, only have a few comments on the pr as everything is pretty clean. Main two things
- Left a tip on how to organize api routes by folders
- Take the agent stuff like agent_bundle.py and contract_schema.py and put them in a new folder instead of leaving them inside of tracker/

services/tracker/src/tracker/agent/contract.py
JarettForzano, line 1: Split the methods in here since some are only used on the client / can be used

services/tracker/src/tracker/config.py
```diff
@@ -80,7 +80,9 @@ async def _init_worker_observability(*_args: object, **_kwargs: object) -> None:
     configure_observability("valkyrie-worker", environment=ENVIRONMENT)
 
 
-# Auth settings
 AUTH_REQUIRED = os.environ.get("AUTH_REQUIRED", "false").lower() == "true"
 DESCOPE_PROJECT_ID = os.environ.get("DESCOPE_PROJECT_ID", "")
 DESCOPE_MANAGEMENT_KEY = os.environ.get("DESCOPE_MANAGEMENT_KEY", "")
+
+CORS_ALLOWED_ORIGINS: list[str] = ["http://localhost:3000"]
+CORS_ALLOWED_ORIGIN_REGEX = r"https://([^/]+\.)?platform\.vals\.ai"
```
JarettForzano: What is the reasoning behind making requests directly from the platform web-app instead of from the BE?

services/tracker/src/tracker/sandbox.py
```diff
@@ -237,7 +242,7 @@ async def upload_agent_artifacts(
     logger.info(f"Uploading contract {contract.name} to sandbox {sandbox.name}")
 
     contract_s3_key = get_benchmark_contract_s3_key(benchmark_id, contract.name)
-    presigned_url = await create_presigned_url(contract_s3_key, aws, s3_bucket)
+    presigned_url = await create_presigned_url(contract_s3_key, aws, s3_bucket, expiration=86400)
```
JarettForzano: expiration can be a const

services/tracker/main.py
```diff
@@ -736,22 +857,25 @@ async def fetch_benchmark_metadata(
...
     harness_config: HarnessConfig = Depends(fetch_harness_config),
     org: Org = Depends(get_current_org),
     task_ids: list[str] | None = Query(default=None),
-) -> StreamingResponse:
+    s3: bool = Query(default=False),
+) -> StreamingResponse | S3UploadResultsResponse:
     """
-    Stream a tar file with agent outputs to the client.
+    Stream a tar file with agent outputs to the client, or upload it to S3 and
+    return a presigned URL when ``?s3=true``.
```
JarettForzano: Hmm, could we prevent loading the entire tar file into memory with s3 multipart upload?

services/tracker/src/tracker/auth.py
```diff
@@ -241,6 +241,63 @@ def forward_tracker_api_key(
...
+    return org
+
+
+def get_current_org(request: Request, session: Session = Depends(get_session)) -> Org:
+    """Resolve the current org from either an Authorization: Bearer or x-api-key header."""
+    if not AUTH_REQUIRED:
+        return get_default_org(session)
+
+    bearer = _extract_bearer_token(request)
+    api_key = request.headers.get("x-api-key")
```
JarettForzano: You can avoid this `assert api_key is not None` by defaulting to a empty string, its still falsy

.gitignore
oalmatov: yes, can remove

services/tracker/src/tracker/config.py
oalmatov: i thought this would save a roundtrip

services/tracker/main.py
oalmatov: im gonna remove downloading all agent outputs feature from the frontend for now and do this properly later. i think this is quite problematic as it stands right now, but want to prioritize getting the feature out.

services/tracker/src/tracker/auth.py
oalmatov: valid

services/tracker/src/tracker/config.py
oalmatov: i think it makes sense to make requests from frontend directly, instead of making a request to platform be then to tracker service. will leave as is

services/tracker/src/tracker/auth.py
oalmatov: Done — `get_current_org` now defaults x-api-key to `""` and drops the assert.

services/tracker/src/tracker/sandbox.py
oalmatov: Done — extracted as `CONTRACT_DOWNLOAD_URL_EXPIRES_SECONDS` in sandbox.py.

services/tracker/main.py
oalmatov: Removed the bulk download-to-S3 (`?s3=true`) feature for now — it built the whole tar in memory. The default streaming path (used by the CLI) stays; we'll re-add the bulk download later backed by a streaming/multipart upload.

services/tracker/src/tracker/agent/contract.py
oalmatov: Done — split into tracker.agent.bundler (client-side zip packaging) and tracker.agent.contract (shared yaml/zip contract resolution).

services/tracker/src/tracker/api/single_benchmark.py
oalmatov: Done — collapsed to a single `.in_(statuses)`.

services/tracker/src/tracker/aws/s3.py
oalmatov: Good point — agents are only stored as `.zip`, so I dropped the subfolder-layout branch; `list_agents` now just lists `agents/<name>.zip`.

tests/unit/test_tracker_service.py
oalmatov: Restored — re-added the secret-override argument + JSON-body assertion to the retry/resume test (the CLI still forwards `secrets`).

services/tracker/src/tracker/api/filter_options.py
oalmatov: Keeping the flat `api/` for now (6 endpoint modules + parsing helper is readable); will revisit a `api/benchmarks/` subpackage as it grows.

services/tracker/src/tracker/api/benchmark_services.py
oalmatov: Keeping the health/error shape — `_ping_service` also handles transport errors (timeout / connection refused) where there's no status code, so status_code alone can't represent an unreachable service.

services/tracker/src/tracker/config.py
oalmatov: Intentional — the platform web-app calls the tracker directly (hence the CORS config + the Bearer-session auth mode); keeping as-is.

.gitignore
oalmatov: Yes — it ignores a local-only `docker-compose.override.yml` for per-dev port remaps (so it's never committed). Keeping it.

PR #489, opened 2026-06-25
tracker: run-level cloudwatch/s3 URLs + sortable task list

oalmatov:
## Run-level console links
`GET /benchmarks/{id}` now returns `cloudwatch_url` + `s3_bucket_url` for the whole run, built from the `X-Harness-*` headers via `get_benchmark_log_url` / `create_benchmark_url` — the same helpers the start-benchmark flow already uses (no new URL logic). Both are null when harness headers are absent (`try_fetch_harness_config`).

## Sortable task list
`GET /benchmarks/{id}/tasks` gains `sort` (`task_id` | `started_at` | `duration` | `status`) and `sort_dir` (`asc`/`desc`). `status` orders by an attention-priority `CASE` so `desc` surfaces errors first (errors → stopped → finished → evaluating → in-progress → building → pending). Default remains newest-first.

Added a test for the status-priority ordering.

Pairs with platform PR vals-ai/platform#2454 (run-link buttons + sortable headers).

### DIFF
--- a/services/tracker/src/tracker/api/single_benchmark.py
+++ b/services/tracker/src/tracker/api/single_benchmark.py
@@ -105,7 +105,7 @@ def get_benchmark_tasks(
     benchmark_id: UUID,
     status: str = Query(default=""),
     task_id_search: str | None = None,
-    sort: Literal["", "task_id", "started_at", "duration", "status"] = Query(default=""),
+    sort: Literal["task_id", "started_at", "duration", "status"] = Query(default="started_at"),
     sort_dir: Literal["asc", "desc"] = Query(default="desc"),
     limit: int = Query(default=50, ge=1, le=500),
     offset: int = Query(default=0, ge=0),
@@ -114,7 +114,7 @@ def get_benchmark_tasks(
 ) -> TasksResponse:
     """Paginated tasks for a benchmark, with optional status filter + task-id search.
 
-    sort=status desc surfaces errors first (attention priority). Default: newest first."""
+    sort=status desc surfaces errors first (attention priority). Default: started_at desc."""
     get_scoped(Benchmark, benchmark_id, session, org)
 
     statuses = parse_csv(status, TaskStatus)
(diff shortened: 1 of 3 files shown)


services/tracker/src/tracker/api/single_benchmark.py
```diff
@@ -65,12 +103,17 @@ def get_benchmark_tasks(
...
     limit: int = Query(default=50, ge=1, le=500),
     offset: int = Query(default=0, ge=0),
     org: Org = Depends(get_current_org),
     session: Session = Depends(get_session),
 ) -> TasksResponse:
-    """Paginated tasks for a benchmark, with optional status filter + task-id search."""
+    """Paginated tasks for a benchmark, with optional status filter + task-id search.
+
+    sort selects the column (task_id | started_at | duration | status); sort_dir is
+    asc | desc. sort=status desc surfaces errors first. Default: newest first."""
```
oalmatov: Can we use Literal here and assign concrete types instead of specifying them in docstring?

oalmatov: Done — `sort` and `sort_dir` are now `Literal`s (FastAPI validates + emits them as an OpenAPI enum); dropped the value list from the docstring. b9df91c6

services/tracker/src/tracker/api/single_benchmark.py
```diff
@@ -65,12 +105,16 @@ def get_benchmark_tasks(
     benchmark_id: UUID,
     status: str = Query(default=""),
     task_id_search: str | None = None,
+    sort: Literal["", "task_id", "started_at", "duration", "status"] = Query(default=""),
```
JarettForzano: Can we use `None` for default sort?

services/tracker/src/tracker/api/single_benchmark.py
```diff
@@ -85,9 +129,21 @@ def get_benchmark_tasks(
...
+    sort_expr = {
+        "task_id": col(Task.task_id),
+        "started_at": col(Task.started_at),
+        "duration": func.coalesce(col(Task.finished_at), func.now()) - col(Task.started_at),
+        "status": _STATUS_SORT_PRIORITY,
+    }.get(sort)
+
+    started_at_desc = col(Task.started_at).desc()
+    if sort_expr is None:
+        order_by = [started_at_desc]
```
JarettForzano: When is sort_expr None?

JarettForzano (approved):
Only a small comment, approving for now but can you swap the default to `None`

JarettForzano (approved):

services/tracker/src/tracker/api/single_benchmark.py
oalmatov: Went with `default="started_at"` instead — started_at desc is already the natural default order, so a concrete field keeps `sort` always-valid and avoids a None/empty branch (which is what your other comment flagged). Let me know if you'd still prefer `None`. (701bf254)

services/tracker/src/tracker/api/single_benchmark.py
oalmatov: No longer applicable — with the Literal default `started_at`, `sort` is always one of the 4 fields, so I dropped the `.get()`/None branch and index the dict directly. sort_expr is never None now. (701bf254)

PR #505, opened 2026-06-30
Add agent egress allowlist support

JarettForzano:
## Summary
Adds an optional agent contract `egress_allowlist` field and uses it to restrict sandbox egress only while the agent command is running. Setup keeps normal network access, then runtime traffic is narrowed through provider-owned egress rules so agents can reach model providers while other outbound requests are denied.

## Changes
- Add `egress_allowlist` to YAML agent contracts and persisted contract requests.
- Apply `sandbox.modify_egress_rules(...)` immediately before `run_cmd` executes.
- Clear egress rules with `sandbox.clear_egress_rules()` after the agent command returns.
- Keep provider-specific URL/domain-to-network-rule conversion inside create-benchmark-service.
- Document the contract field and update focused unit coverage.
- Pin tracker to the create-benchmark-service commit that resolves Daytona allowlist hosts.

## Related Issues
Depends on vals-ai/create-benchmark-service#74

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor
- [x] Docs / config

## Testing
- [x] Manually tested

Validation run:
- Focused tracker sandbox unit coverage.
- Live tracker sandbox egress integration against the pinned CBS commit.
- Lint, formatting, and static type checks on touched tracker files.

## Checklist
- [x] Self-reviewed the diff
- [x] Docs updated if needed


services/tracker/src/tracker/sandbox.py
```diff
@@ -595,10 +595,18 @@ async def run_agent(
     # Create cwd if it does not already exist
     await _exec(sandbox, f"mkdir -p {shlex.quote(cwd)}")
 
+    if contract.egress_allowlist:
+        await sandbox.modify_egress_rules(contract.egress_allowlist)
```
tthuwng: Two issues with this call sitting outside the `try`/`finally` below:

1. **Partial application isn't cleaned up.** `modify_egress_rules` writes the `/etc/hosts` pins and then calls `update_network_settings`; if it raises after the restriction has (partially) applied — e.g. retries exhaust on a flaky Daytona API — we propagate before entering the `try`, the `finally` never runs, and the sandbox stays locked down for everything that follows. Moving the apply inside the `try` (or into a small `asynccontextmanager` that owns both apply and clear — `asynccontextmanager` is already imported here) closes that gap and also collapses the doubled `if contract.egress_allowlist:` guards.
2. **Error taxonomy.** `_resolve_daytona_allowed_addresses` raises raw `ValueError` (e.g. a host that transiently fails DNS on the tracker host), which bypasses the tracker `SandboxError` taxonomy — `process_task` retries only `SandboxSetupError`, so a DNS blip becomes an unclassified task failure instead of a retryable setup error. Mapping these through the same error translation the other sandbox calls use (`_exec`-style) would keep the taxonomy intact.

services/tracker/src/tracker/sandbox.py
```diff
@@ -595,10 +595,18 @@ async def run_agent(
...
-    exit_reason, agent_run_time = await stream_command_output(
-        sandbox, f"cd {shlex.quote(cwd)} && PYTHONSAFEPATH=1 {run_cmd}", log_output
-    )
+    try:
+        exit_reason, agent_run_time = await stream_command_output(
+            sandbox, f"cd {shlex.quote(cwd)} && PYTHONSAFEPATH=1 {run_cmd}", log_output
+        )
+    finally:
+        if contract.egress_allowlist:
+            with suppress(Exception):
```
tthuwng: `suppress(Exception)` with no logging is risky here: if `clear_egress_rules` fails (it can, even with retries — Daytona API blip, rate limit), `run_agent` returns normally and `evaluate_instance` then runs its eval commands **in the same still-restricted sandbox**. Anything network-dependent in evaluation fails or mis-scores with no diagnostic pointing at stale egress rules.

At minimum, log the swallowed exception (`logger.warning("failed to clear egress rules for sandbox %s", sandbox.id, exc_info=True)`); better, surface it so the task can be marked infra-failed rather than silently mis-scored. Also note the clear resets to fully unrestricted rather than restoring whatever network settings the sandbox was created with — fine today, but worth a comment since it would silently wipe any future baseline restriction.

services/tracker/pyproject.toml
```diff
@@ -51,7 +51,7 @@ build-backend = "hatchling.build"
 prerelease = "allow"
 
 [tool.uv.sources]
-create-benchmark-service = { git = "https://github.com/vals-ai/create-benchmark-service.git", tag = "v0.8.2" }
+create-benchmark-service = { git = "https://github.com/vals-ai/create-benchmark-service.git", rev = "80fc0b22c3a4ddccb957bfe5b4b390adec033a1b" }
```
tthuwng: This rev (`80fc0b22`) is the head of the still-open CBS #74 branch (`jf/network-blocking`), which has diverged from CBS main (ahead 6 / behind 3) — so merging this PR as-is ships the tracker against unreviewed branch code that's also missing three main-only commits. Suggest: land CBS #74 first, cut a release tag, and repoint this to the tag before merging (also avoids depending on an unmerged-branch commit staying fetchable after the branch is deleted).

docs/CONTRACTS.md
```diff
@@ -169,6 +173,18 @@ benchmarks/run_id/task_0/artifacts/config.json
 benchmarks/run_id/task_0/artifacts/result.json
 ```
 
+### `egress_allowlist: list`
+
+URLs the agent may reach while `run_cmd` is running. Use this to allow model provider requests while denying other outbound requests from the agent sandbox; the sandbox provider resolves each host into its network rules at run time.
```
tthuwng, line 178: Suggest stating the scope of the guarantee here: the restriction is temporal, not a hard security boundary. A background process the agent leaves running gets unrestricted egress the moment rules clear after `run_cmd`; the agent (root) can rewrite `/etc/hosts`; and allowing a CDN-fronted host allowlists its edge IPs, which serve many other origins. That's fine for eval hygiene (keeping agents on their model providers), but benchmark authors shouldn't read this as exfiltration-proof containment.

tthuwng (commented):
Reviewed together with vals-ai/create-benchmark-service#74. The contract field + run-window scoping design is clean and the unit/integration coverage is solid. Inline comments below; two things that don't anchor to the diff:

**Older CLIs silently drop the allowlist.** `start_benchmark` only re-resolves the contract from S3 when both `install_cmd` and `run_cmd` are empty (`main.py:289`), and the standard valk CLI path builds `AgentContractRequest` client-side with `run_cmd` populated — so any installed CLI predating `egress_allowlist` serializes contracts without the field and the agent runs with unrestricted egress even though its YAML declares an allowlist, with no warning. Worth deciding how to handle before this becomes a trust boundary: server-side re-resolution of the allowlist field, or a CLI version gate.

**Merge sequencing.** This should land after CBS #74 merges and gets a release tag, with the pin repointed (see pyproject comment).

JarettForzano:
> Reviewed together with [vals-ai/create-benchmark-service#74](https://github.com/vals-ai/create-benchmark-service/pull/74). The contract field + run-window scoping design is clean and the unit/integration coverage is solid. Inline comments below; two things that don't anchor to the diff:
> 
> **Older CLIs silently drop the allowlist.** `start_benchmark` only re-resolves the contract from S3 when both `install_cmd` and `run_cmd` are empty (`main.py:289`), and the standard valk CLI path builds `AgentContractRequest` client-side with `run_cmd` populated — so any installed CLI predating `egress_allowlist` serializes contracts without the field and the agent runs with unrestricted egress even though its YAML declares an allowlist, with no warning. Worth deciding how to handle before this becomes a trust boundary: server-side re-resolution of the allowlist field, or a CLI version gate.
> 
> **Merge sequencing.** This should land after CBS #74 merges and gets a release tag, with the pin repointed (see pyproject comment).

Should not need to worry about runs that are in progress

services/tracker/pyproject.toml
JarettForzano: Will update when CBS is merged

tthuwng:
This is not about runs already in progress. It affects any new run started after the tracker deploy by an older installed `valk`: that client parses the S3 contract with a schema that does not know `egress_allowlist`, then sends populated `install_cmd`/`run_cmd`. The new server therefore skips `_resolve_contract_from_s3`, Pydantic defaults the missing field to `[]`, and the agent runs with unrestricted egress even though its YAML declares an allowlist.

Please either overlay the authoritative server-side `egress_allowlist` when accepting the request or reject clients that cannot carry the field, and add a regression test using a legacy payload with populated commands.

tthuwng (approved):

JarettForzano:
> This is not about runs already in progress. It affects any new run started after the tracker deploy by an older installed `valk`: that client parses the S3 contract with a schema that does not know `egress_allowlist`, then sends populated `install_cmd`/`run_cmd`. The new server therefore skips `_resolve_contract_from_s3`, Pydantic defaults the missing field to `[]`, and the agent runs with unrestricted egress even though its YAML declares an allowlist.
> 
> Please either overlay the authoritative server-side `egress_allowlist` when accepting the request or reject clients that cannot carry the field, and add a regression test using a legacy payload with populated commands.

I don't think we need to worry about that since we dont have anyone who is running valkyrie locally right now

PR #521, opened 2026-07-09
Guard task generations across stop and retry

tthuwng:
## Summary

- attach exact task attempt timestamps to new start and retry queue messages
- reject delayed messages and stale task writes after a stop and retry
- cancel old task coroutines when their task generation changes
- revalidate the full task snapshot before committing a final score
- keep legacy list messages readable for a two-phase rollout
- expose the default-off producer flag through the tracker task definition and GitHub Actions variable

## Verification

- tracker unit suite: 223 passed
- focused tracker suite: 57 passed
- infra suite: 9 passed and 9 subtests passed
- targeted basedpyright: 0 errors
- Ruff, formatting, and git diff checks pass

## Required rollout

1. Deploy this reader-compatible code with TOKENIZED_RETRY_BATCHES_ENABLED=false.
2. Force-stop target runs and fully drain old worker revisions, legacy queue messages, sandboxes, and uploads.
3. Set the repository Actions variable TOKENIZED_RETRY_BATCHES_ENABLED=true and deploy again.
4. Only then enqueue Luna retries.

## Explicit boundaries

This PR prevents stale database state adoption. It does not add leases for hard worker death, and it does not make fixed S3 task artifacts or final-view Lambdas generation-safe. The FAB watcher must remain stopped until the previous finalizer is known quiescent. Durable artifact publication needs generation-scoped staging plus conditional promotion in a separate change.

No deployment is included in this PR.

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -77,6 +77,7 @@ jobs:
           AUTH_REQUIRED: ${{ secrets.AUTH_REQUIRED }}
           BENCHMARK_CATALOG_URL: ${{ secrets.BENCHMARK_CATALOG_URL }}
           DESCOPE_PROJECT_ID: ${{ secrets.DESCOPE_PROJECT_ID }}
+          TOKENIZED_RETRY_BATCHES_ENABLED: ${{ vars.TOKENIZED_RETRY_BATCHES_ENABLED }}
         run: |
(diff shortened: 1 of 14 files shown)


tthuwng:
Closing per release-ops direction. We are keeping this effort focused on running and recovering benchmarks; no deployment was made.

PR #524, opened 2026-07-10
Add provider-generic sandbox orphan cleanup schedule

conjfrnk:
## Summary

- add a provider-neutral cleanup engine over normalized candidate metadata and `list`, `refresh`, and `delete` backend operations
- keep Daytona-specific target inventory, timestamp parsing, transient retry handling, and PAUSED deletion inside a Daytona adapter
- sweep in-scope sandboxes strictly older than 48 hours and exempt only an exact `clean-up` key whose trimmed, case-insensitive value is `false`
- finish inventory before mutation, then refresh and revalidate identity, scope, age, and opt-out metadata immediately before each deletion
- run hourly in production through generic `SANDBOX_CLEANUP_*` configuration and generic Lambda, schedule, log, and DLQ resource names

## Provider contract

The policy engine contains no Daytona types or lifecycle branches. A sandbox provider becomes cleanup-capable by supplying normalized candidate metadata and implementing complete-list, refresh, and delete operations. Daytona is the first and currently configured adapter. Unsupported providers fail before listing or deleting anything; this PR does not pretend Modal has the age/tag inventory needed for a safe implementation.

## Safety

- the production schedule is disabled unless `SANDBOX_CLEANUP_ENABLED=true`
- deletion remains dry-run unless `SANDBOX_CLEANUP_DRY_RUN=false`; missing or unknown rollout values fail closed
- the configured provider and secret are explicit, while Daytona remains the safe default using `AgenticHarnessSecrets`
- provider inventory completes before mutation; each delete is bounded to 120 seconds; individual failures are recorded while later candidates continue
- scope mismatches, invalid timestamps, mismatched refreshed identities, and unsupported providers make the invocation fail
- cleanup resources and IAM are production-only and narrowly scoped; the Lambda does not inherit tracker database, Redis, S3, broker, or Sentry configuration

## Verification at `d1057788`

- cleanup unit suite: 50 cases, 100% statement and branch coverage across the generic engine and Daytona adapter
- tracker unit/local integration suite: 453 passed, 87.17% coverage
- CLI unit/local integration suite: 197 passed
- infrastructure suite: 22 passed; infrastructure lint and type checking passed
- root, tracker, and infrastructure formatting/lint passed; root and changed tracker cleanup files/tests type-check with zero errors
- both frozen lock checks passed and resolve `create-benchmark-service==0.16.3`
- final ARM64 Lambda image built successfully and imported `tracker.sandbox_cleanup.lambda_handler` in-container
- the existing non-mutating live-test safety guard passed
- an independent exact-tree review found no blocking correctness, safety, or regression issue
- tests remain consolidated into one renamed unit file and the existing Daytona live test; no additional test files were added

No AWS deployment, Lambda invocation, schedule enablement, production cleanup, or provider-calling live test was performed for this revision. The exact-head live workflow remains an explicit approval gate.

## Before merge

1. Let the hosted non-live checks finish at exact head `d1057788`.
2. With explicit approval, rerun the tracker live-integration workflow at that exact head.
3. Obtain the required Valkyrie team approval.

## Rollout

Validate the selected provider scope and opt-outs, deploy disabled and dry-run, inspect logs and the DLQ, then require explicit approval before enabling deletion.


JarettForzano (changes requested):
Make it generic so we can use it with any sandbox provider

conjfrnk:
Addressed in `0a5f6233`: the cleanup policy now runs against a provider-neutral candidate/backend contract, while Daytona target inventory, timestamp parsing, retries, and PAUSED deletion are isolated in the Daytona adapter. The Lambda/config/resources are now `SANDBOX_CLEANUP_*`; unsupported providers fail before mutation. Tests stay in one renamed unit file plus the existing Daytona live test—no additional test files. @JarettForzano, could you please take another look?

services/tracker/src/tracker/daytona_cleanup.py
JarettForzano, line 1: this file is still daytona specific

services/tracker/src/tracker/daytona_cleanup.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def list(self, query: ListSandboxesQuery | None = None) -> AsyncIterator[AsyncSandbox]: ...
+
+    async def get(self, sandbox_id_or_name: str) -> AsyncSandbox: ...
+
+
+class SandboxDeleteProvider(Protocol):
+    """Existing sandbox-provider deletion operation used by the Daytona adapter."""
+
+    async def delete_sandbox(self, instance_id: str) -> None: ...
```
JarettForzano: for some background protocol is usually what people use they want to enforce that the methods exist inside of the object they are calling. If you inherit the class its an explicit protocol (all methods must exist before the object is initialized), if its just used implicitly, it just needs the types that are inside of the protocol and does not need to inherit the protocol at all.

```python
class Example(Protocol):
    def read(self) -> str: ...
    def close(self) -> None: ...


class Reader:
    def read(self) -> str:
        return "data"


def process(value: Example) -> None:
    value.read()

process(Reader())  # Type-checking error: close() is missing
```
Here we know that the methods exist because we created the class, so these are not needed. Instead you can just utilize the sandbox provider object from the CBS package. In this case you can say we explicitly create the class with the correct methods.

services/tracker/pyproject.toml
```diff
@@ -25,7 +25,8 @@ dependencies = [
     "tenacity>=9.0.0",
     "sentry-sdk[fastapi]==2.58.0",
     "python-json-logger>=3.2.1",
-    "create-benchmark-service>=0.11.0",
+    "create-benchmark-service>=0.16.3",
+    "daytona>=0.189.0",
```
JarettForzano: Can access daytona through the CBS package

JarettForzano (changes requested):
`services/tracker/src/tracker/daytona_cleanup.py` and `services/tracker/src/tracker/sandbox_cleanup.py `?

services/tracker/src/tracker/sandbox_cleanup.py
```diff
@@ -0,0 +1,394 @@
...
+    return value is not None and value.strip().casefold() == _CLEANUP_DISABLED
+
+
+def _candidate_exclusion(
+    candidate: CleanupCandidate,
+    *,
+    scope: str | None,
+    cutoff: datetime,
+) -> CandidateExclusion | None:
+    if candidate.scope != scope:
```
JarettForzano: Where are these set and what do the different options do?

services/tracker/src/tracker/sandbox_cleanup.py
JarettForzano, line 1: remove CleanupCandidate, SandboxCleanupBackend, SecretsManagerClient, and the provider protocols. Use the CBS Sandbox/SandboxProvider types

services/tracker/src/tracker/sandbox_cleanup.py
```diff
@@ -0,0 +1,394 @@
...
+        )
+    elif exclusion == "invalid_metadata":
+        logger.error(
+            "Skipping sandbox with invalid creation metadata",
+            extra={"provider": provider, "sandbox_id": candidate.id, "sandbox_name": candidate.name},
+        )
+
+
+async def cleanup_old_sandboxes(
+    backend: SandboxCleanupBackend,
```
JarettForzano: [Codex Comment]: the main loop only needs list, classify, refresh, reclassify, delete, and increment an outcome. Preserve those safeguards, but return the counter directly instead of constructing several intermediate objects.

services/tracker/src/tracker/sandbox_cleanup.py
```diff
@@ -0,0 +1,394 @@
...
+    if not isinstance(parsed, dict):
+        raise RuntimeError("Sandbox cleanup secret must contain a JSON object")
+    payload = cast(dict[str, object], parsed)
+
+    try:
+        return sandbox_provider_config_from_mapping({**payload, "type": provider_type})
+    except (TypeError, ValueError):
+        # Validation details may echo provider credentials, so expose only the
+        # provider discriminator and keep the original exception unchained.
+        raise RuntimeError(f"Sandbox cleanup secret is invalid for provider {provider_type!r}") from None
```
JarettForzano, line 161: reuse the tracker’s existing provider-secret loader instead of maintaining another

services/tracker/src/tracker/sandbox_cleanup.py
```diff
@@ -0,0 +1,394 @@
...
+@asynccontextmanager
+async def _cleanup_backend(provider_config: SandboxProviderConfig) -> AsyncGenerator[SandboxCleanupBackend]:
+    if provider_config.type == "daytona":
+        from tracker.daytona_cleanup import daytona_cleanup_backend
+
+        async with daytona_cleanup_backend(provider_config) as backend:
+            yield backend
+        return
+
+    raise RuntimeError(f"Sandbox cleanup does not support provider {provider_config.type!r}")
```
JarettForzano: Construct the CBS provider from SandboxProviderConfig and pass it directly to the reaper

.github/workflows/deploy.yaml
```diff
@@ -83,6 +83,10 @@ jobs:
           AUTH_REQUIRED: ${{ secrets.AUTH_REQUIRED }}
           BENCHMARK_CATALOG_URL: ${{ secrets.BENCHMARK_CATALOG_URL }}
           DESCOPE_PROJECT_ID: ${{ secrets.DESCOPE_PROJECT_ID }}
+          SANDBOX_CLEANUP_ENABLED: ${{ vars.SANDBOX_CLEANUP_ENABLED }}
+          SANDBOX_CLEANUP_DRY_RUN: ${{ vars.SANDBOX_CLEANUP_DRY_RUN }}
```
JarettForzano: I don't think its needed to add a dry run var to the deploy workflow

infra/README.md
JarettForzano, line 1: You can have devin trim this some more to make it less verbose

JarettForzano (changes requested):
I would say trim the unit tests and add one nice integration test

services/tracker/pyproject.toml
```diff
@@ -50,7 +50,7 @@ build-backend = "hatchling.build"
 prerelease = "allow"
 
 [tool.uv.sources]
-create-benchmark-service = { git = "https://github.com/vals-ai/create-benchmark-service.git", rev = "v0.20.0" }
+create-benchmark-service = { git = "https://github.com/vals-ai/create-benchmark-service.git", rev = "2b491d07b0f7e509a2703d42689b5f9214e54a0d" }
```
JarettForzano: update to latest release

JarettForzano (approved):

PR #532, opened 2026-07-10
refactor: add aws client providers

BradenBug:
### Why

Before managed mode can exist, the tracker needs one place where AWS credentials get chosen. Today every AWS call site takes loose `aws`/`s3_bucket`/`region` parameters, so there is nowhere to plug in a second credential source. Credential-bearing runtime objects can also show up in generic reprs or instrumented arguments. This PR builds the provider boundary and hardens those specific telemetry paths.

---

AWS credentials are now selected behind one runtime boundary before tracker and CLI operations reach AWS. Existing access-key runs still build the same clients, while the SDK credential chain is available for deployment roles without passing credentials through application payloads.

## Stack

1. **[#532 — aws client providers](https://github.com/vals-ai/Valkyrie/pull/532) ← this PR**
2. [#543 — persist managed aws runtime](https://github.com/vals-ai/Valkyrie/pull/543)
3. [#544 — managed taskiq execution](https://github.com/vals-ai/Valkyrie/pull/544)

## What changed

- `AWSClientProvider` selects either explicit access keys or the AWS SDK credential chain. Each provider owns its presigned-URL lifetime limit.
- `AWSRuntime` carries AWS resource locations and the selected client provider through tracker AWS operations.
- HTTP endpoints receive an `AWSRuntime` dependency at the request boundary. Taskiq workers construct it from the queued execution mode.
- CLI S3 operations use the same runtime and fail with a product-facing configuration error when the bucket or region is missing.
- Credential-bearing fields stay out of model representations, queued-request validation errors, and instrumented function arguments.
- Annotation-only AWS model imports do not load tracker configuration at runtime. Sealed control tasks can install deployment database settings before the database engine is created.

## How to review

Start with `services/tracker/src/tracker/aws/clients.py` and `runtime.py`, then follow one S3 call through `api/dependencies.py`. The important compatibility property is that `AWSRuntime.from_harness_config` still selects the explicit-key provider.

The CLI adapter is `src/valkyrie/cli/s3_config.py`. It may use the SDK credential chain locally, but it still requires an explicit region because the runtime also builds regional console links.

## Testing

Live-tested on dev in the combined stack at `17a16fd6`. A one-task `legal-research` run finished through the deployment credential chain, and the same task finished with explicit harness access-key headers. The sealed release-control task loaded its database configuration, activated protocol-2 artifact `git-17a16fd65659-08c12ea42e897db9`, and completed the maintenance cycle.

Design: architectural (introduces the credential-provider and runtime boundary used by the rest of the stack)


miyukimatthews (approved):

JarettForzano (changes requested):
Accept this meme I had chatgpt generate
Image

services/tracker/src/tracker/utils/run_orchestration.py
```diff
@@ -224,22 +225,40 @@ async def finalize_all_error_run(benchmark_id: UUID, org: Org) -> bool:
         return False
 
 
+def _parse_start_benchmark_request(payload: dict[str, Any]) -> StartBenchmarkRequest:
+    request: StartBenchmarkRequest | None
```
JarettForzano, line 261: Ah i see, so we have a special wrapper to prevent exposing AWS secrets on serialization? Can you make it a doc string so someone does not refactor in the future?

services/tracker/src/tracker/aws/runtime.py
JarettForzano, line 1: The betrayal is real, you swapped AWS to Aws

services/tracker/src/tracker/aws/runtime.py
```diff
@@ -0,0 +1,37 @@
...
+@dataclass(frozen=True)
+class AwsResources:
+    region: str
+    s3_bucket: str
+    log_group: str
+    log_retention_days: int
+
+
+@dataclass(frozen=True)
+class AwsRuntime:
```
JarettForzano: AwsRuntime.managed is not used

services/tracker/src/tracker/_lambda.py
```diff
@@ -1,51 +1,45 @@
...
-        aws_access_key_id=aws.aws_access_key_id,
-        aws_secret_access_key=aws.aws_secret_access_key,
-        aws_session_token=aws.aws_session_token,
-        region_name=aws.aws_default_region,
-        config=config,
-    )
+def _response_status(payload: Any) -> Any:
+    if not isinstance(payload, dict):
+        return None
+    return cast(dict[str, Any], payload).get("statusCode")
```
JarettForzano: Cast but still return Any

src/valkyrie/cli/s3_config.py
JarettForzano, line 1: Why duplicate the tracker helpers?

JarettForzano (changes requested):
Use test generation skill

services/tracker/src/tracker/aws/clients.py
JarettForzano, line 1: Consolidate shared methods into the inherited class

JarettForzano (approved):
Approved, try to reuse fixtures:

```python
   aws_runtime = AWSRuntime.from_harness_config(
        HarnessConfig(
            aws=AWSCredentials(
                aws_access_key_id="test",
                aws_secret_access_key="test",
                aws_default_region="us-east-1",
            ),
            s3_bucket="unused",
            log_group="unused",
            log_retention_policy=1,
            sandbox_provider_secret_name="unused",
```

```python
@pytest.fixture
def aws_runtime(harness_config: HarnessConfig) -> AWSRuntime:
    return AWSRuntime.from_harness_config(harness_config)
```

services/tracker/src/tracker/aws/clients.py
```diff
@@ -0,0 +1,109 @@
...
+    @lru_cache(maxsize=32)
+    def secretsmanager_client(self) -> Any:
+        return _boto3_client("secretsmanager", **self._client_kwargs())
+
+    @lru_cache(maxsize=32)
+    def lambda_client(self, config: Config | None = None) -> Any:
+        return _boto3_client("lambda", config=config, **self._client_kwargs())
+
+    def maximum_presign_ttl(self, requested_seconds: int) -> int:
+        return requested_seconds
```
JarettForzano: These are serving the same purpose right now, the base provider class is inherited so the protocol is going to be explicitly followed. Meaning we probably don't need the protocol at all.

services/tracker/src/tracker/aws/clients.py
```diff
@@ -0,0 +1,109 @@
...
+        return {
+            "aws_access_key_id": self.credentials.aws_access_key_id,
+            "aws_secret_access_key": self.credentials.aws_secret_access_key,
+            "aws_session_token": self.credentials.aws_session_token,
+            "region_name": self.credentials.aws_default_region,
+        }
+
+
+@dataclass(frozen=True)
+class DefaultChainAWSClientProvider(_BaseAWSClientProvider):
```
JarettForzano: I could not find any references to this, are we using it in a future PR?

services/tracker/main.py
```diff
@@ -131,7 +132,20 @@ def _operation_id(route: APIRoute) -> str:
 
 app = FastAPI(generate_unique_id_function=_operation_id, redirect_slashes=False)
 
-logfire.instrument_fastapi(app, excluded_urls="/health$")
+
+def _exclude_request_arguments_from_telemetry(
+    _request: Request | WebSocket,
+    _attributes: dict[str, Any],
```
JarettForzano: Does this exclude more than just credentials?

services/tracker/src/tracker/observability/sentry.py
```diff
@@ -70,6 +70,8 @@ def init_sentry(service_name: str, environment: str) -> None:
             instrumenter=INSTRUMENTER.OTEL,
             enable_logs=True,
             send_default_pii=False,
+            max_request_body_size="never",
+            include_local_variables=False,
```
JarettForzano: Same with this, does this also affect more than just credentials?

services/tracker/src/tracker/utils/run_orchestration.py
BradenBug: done

PR #543, opened 2026-07-13
feat: persist managed aws runtime and route run access

BradenBug:
### Why

Run operations (fetch, analyze, stop, results, artifacts) re-read the `X-Harness-*` headers on every request to build AWS credentials. That breaks for managed runs, which won't send those headers — and a run shouldn't be able to change credentials mid-life anyway. So each run now records which mode it uses, and every later operation reads that record. Nothing can create a managed run yet; that comes later in the stack.

---

Each run now records whether AWS authority comes from caller-provided access keys or the deployment role. Later requests resolve the runtime from that stored mode, so a run cannot silently change credentials during its lifetime.

## Stack

1. [#532 — aws client providers](https://github.com/vals-ai/Valkyrie/pull/532)
2. **[#543 — persist managed aws runtime](https://github.com/vals-ai/Valkyrie/pull/543) ← this PR**
3. [#544 — managed taskiq execution](https://github.com/vals-ai/Valkyrie/pull/544)

## What changed

- Each benchmark row stores `aws_managed`, defaulting to false for existing and access-key runs.
- Consumer-named resolver functions apply the correct policy for run operations, start requests, optional metadata, and agent-library access.
- `GET /aws-runtime` returns deployment-owned resource locations without returning credentials.
- Presigned responses report the provider's actual lifetime. Older tracker responses that omit `expires_in` still parse as the historical one-day lifetime, while the CLI displays values as readable days, hours, minutes, or seconds.
- Force-stopping a managed run with no stored sandbox-provider secret reports the missing configuration instead of asking for access-key headers.

## How to review

Start with `services/tracker/src/tracker/aws/resolver.py`. Each public resolver is named for its consumer, and the stored run mode is authoritative after creation.

Then read the `stop_benchmark` path in `services/tracker/main.py`. A managed run has no access-key harness configuration, so the missing provider-secret branch must fail before sandbox cleanup begins.

## Testing

Live-tested on dev in the combined stack at `17a16fd6`. The no-header managed run remained readable with only the API key through start, completion, and recovery. The explicit-header access-key control remained readable with its harness headers and completed on the same Tracker and executor release.

Design: architectural (persists the AWS authority mode and centralizes runtime selection for HTTP consumers)


miyukimatthews (dismissed):

services/tracker/main.py
```diff
@@ -691,14 +741,25 @@ async def stop_benchmark(
...
+        aws_runtime = runtime_resolution.runtime
+        resolved_harness_config = runtime_resolution.legacy_harness_config
+        provider_secret_name = benchmark_row.arguments.sandbox_provider_secret_name or (
+            resolved_harness_config.sandbox_provider_secret_name if resolved_harness_config is not None else None
         )
+        if not provider_secret_name:
+            detail = "The run does not have a sandbox provider secret name."
+            if resolved_harness_config is not None:
+                detail += " Provide the x-harness-sandbox-provider-secret-name header and retry."
+            raise HTTPException(status_code=400, detail=detail)
```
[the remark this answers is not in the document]
BradenBug: Fixed in 71ec967d. Provider-secret validation now runs before initiate_stop_benchmark, so a rejected force-stop request cannot persist STOPPING/STOPPED state. The regression test reloads the run and pending task from the database, verifies both remain unchanged, and verifies sandbox cleanup is not called.

JarettForzano (commented):
When we say legacy, are we always referencing the prev way with credentials? If we don't plan to deprecate it  I would prefer we do not call it legacy. Since that makes me think we are going to deprecate it down the line.

JarettForzano (changes requested):
From what I see it looks like there is a simplification on how we are passing the harness config to the service.

The request body contains the very same information we are using to make `OptionalHarnessConfig` and send to the server. You can remove legacy_harness_config and instead just pass the request and validate server-side.

```python
- aws_runtime = resolve_non_run_aws_runtime(request, org.id, legacy_harness_config)
+ aws_runtime = resolve_non_run_aws_runtime(request, org.id)
```

services/tracker/src/tracker/utils/harness_config.py
JarettForzano, line 1: Can you make a note to deprecate this?

services/tracker/main.py
```diff
@@ -697,19 +747,34 @@ async def stop_benchmark(
...
-            benchmark_row.arguments.sandbox_provider_secret_name or harness_config.sandbox_provider_secret_name
+        resolved_harness_config = runtime_resolution.legacy_harness_config
+        provider_secret_name = benchmark_row.arguments.sandbox_provider_secret_name or (
+            resolved_harness_config.sandbox_provider_secret_name if resolved_harness_config is not None else None
         )
+        if not provider_secret_name:
+            detail = "The run does not have a sandbox provider secret name."
+            if resolved_harness_config is not None:
+                detail += " Provide the x-harness-sandbox-provider-secret-name header and retry."
+            raise HTTPException(status_code=400, detail=detail)
```
JarettForzano: This is a bit dense, could we simplify it?

services/tracker/main.py
JarettForzano, line 1: There is a common pattern with the AWS initialization process, I think to simplify we should make it a depends and then available by default through the route. This is optional feedback since it targets behavior that existed before your changes. If you want to you can clean this up.

```python
def get_run_with_aws(
    benchmark_id: TrackedBenchmarkId,
    request: Request,
    harness_config: HarnessConfig | None = Depends(try_fetch_harness_config),
) -> tuple[Benchmark, AWSRuntime]:
    benchmark = get_scoped(Benchmark, benchmark_id, session, org)
    runtime = resolve_run_aws_runtime(
        request,
        aws_managed=benchmark.aws_managed,
        org_id=org.id,
        legacy_harness_config=harness_config,
    ).runtime

    return benchmark, runtime


AuthenticatedRun = Annotated[
    tuple[Benchmark, AWSRuntime],
    Depends(get_run_with_aws),
]
```

services/tracker/src/tracker/aws/resolver.py
JarettForzano, line 1: Can we add docstrings on all of these?

JarettForzano (approved):

PR #544, opened 2026-07-13
feat: add managed taskiq execution

BradenBug:
### Why

Earlier layers store which mode a run uses. This PR makes the execution path honor it: starting, resuming, and running a benchmark. Before, every queue message and worker path assumed access keys were inside. After, a managed run's queue message carries no credentials at all — the worker uses its own task role.

---

Managed starts and resumes now reach the executor without carrying AWS credentials. The stable ExecutorHost accepts both the existing access-key message and the new managed message, while release admission prevents managed work from reaching an executor artifact that cannot understand it.

## Stack

1. [#532 — aws client providers](https://github.com/vals-ai/Valkyrie/pull/532)
2. [#543 — persist managed aws runtime](https://github.com/vals-ai/Valkyrie/pull/543)
3. **[#544 — managed taskiq execution](https://github.com/vals-ai/Valkyrie/pull/544) ← this PR**

## What changed

**Start and resume**

- Managed requests reject AWS credential fields recursively and require a sandbox provider and stored provider-secret name.
- Resume reconstructs the request from the stored run mode. A run cannot switch authority modes.
- A running benchmark rejects secret overrides unless the request sets `retry=true`; overrides are never accepted and silently discarded.
- Queue payload construction happens before the database commit. A deterministic serialization failure rolls back the benchmark, tasks, and dispatch instead of leaving work marked in progress with no queue message.
- Start admission copies the agent artifact with its S3 version identity. Cleanup deletes that copy only when a fresh database read proves the benchmark and dispatch were not committed.

**Queue and execution**

- Access-key runs retain the protocol-1 message fields.
- Managed runs use a credential-free protocol-2 `ManagedExecutionContext`.
- The stable ExecutorHost normalizes either shape once and launches the selected immutable executor artifact.
- Managed start and recovery admission require an active protocol-2 release. Protocol-1 releases and already-queued access-key work continue to drain.
- Managed execution checks CloudWatch Logs, Secrets Manager, and Lambda permissions before creating a sandbox.

**Rollout**

- `AWS_MANAGED_SUBMISSIONS_ENABLED` defaults to false and gates new managed starts. Existing managed work can still resume or drain after the gate closes.

## How to review

Read the boundary in this order:

1. `services/tracker/src/tracker/types.py` — credential-free managed request and envelope validation.
2. `services/tracker/main.py` and `tracker/executor/dispatch_control.py` — transaction, release admission, rollback, and enqueue ownership.
3. `services/executor_host/supervisor.py` and `executor_protocol.py` — protocol-1/protocol-2 normalization.
4. `services/tracker/src/tracker/utils/run_orchestration.py` — `_parse_queued_execution` and `_preflight_managed_aws`.

```mermaid
flowchart LR
    Client --> Tracker
    Tracker -->|benchmark and dispatch| DB[(Database)]
    Tracker -->|versioned agent copy| S3[(S3)]
    Tracker -->|V1 access keys or V2 managed context| Redis
    Redis --> Host[Stable ExecutorHost]
    Host --> Artifact[Versioned executor artifact]
    Release[Active release protocol] --> Tracker
    Tracker -. delete copy only after rollback and proven DB absence .-> S3
```

## Testing

Live-tested on dev in the combined stack at `17a16fd6`. A no-header one-task `legal-research` run finished through Taskiq protocol 2, the stable ExecutorHost, Linux ARM artifact `git-17a16fd65659-08c12ea42e897db9`, the deployment credential chain, S3, CloudWatch Logs, and the Daytona secret. A managed retry was accepted and re-dispatched under the same release. The explicit-header compatibility control finished through the access-key message path.

Design: architectural (adds a versioned queue protocol, executor-release compatibility gate, and DB/S3 ownership boundary)


miyukimatthews (dismissed):

services/tracker/src/tracker/aws/resolver.py
```diff
@@ -130,6 +130,19 @@ def fetch_harness_config(request: Request) -> HarnessConfig:
     _raise_missing_header(state.first_missing_key)
 
 
+def resolve_start_harness_config(request: Request, body_config: HarnessConfig | None) -> HarnessConfig | None:
+    """Apply access-key header-over-body precedence for a start request."""
+    state = inspect_harness_headers(request)
```
JarettForzano: can we make "state" more specific, i dont get what it is unless i look at the method that creates it?

services/tracker/src/tracker/middleware/logging_context.py
```diff
@@ -15,9 +15,14 @@ async def pre_execute(self, message: TaskiqMessage) -> TaskiqMessage:
         benchmark_id_var.set("")
         task_id_var.set("")
 
-        benchmark_id_str = message.kwargs.get("benchmark_id_str", "")
+        message_kwargs = message.kwargs
+        benchmark_id_str = message_kwargs.get("benchmark_id_str", "")
+        if not benchmark_id_str:
+            execution_context = message_kwargs.get("execution_context_json")
+            if isinstance(execution_context, dict):
+                benchmark_id_str = cast(dict[str, Any], execution_context).get("benchmark_id", "")
```
JarettForzano: this can be consolidated in a helper method

services/tracker/src/tracker/utils/resources.py
```diff
@@ -66,13 +66,27 @@ def create_benchmark_service_client_from_request(request: StartBenchmarkRequest)
...
+    aws_managed: bool,
+) -> Benchmark:
     """Convert a StartBenchmarkRequest to a Benchmark database model."""
+    if aws_managed != (request.harness_config is None):
+        raise ValueError("Benchmark AWS mode does not match the start request")
+    provider_secret_name = (
+        request.harness_config.sandbox_provider_secret_name if request.harness_config is not None else None
+    ) or request.sandbox_provider_secret_name
+    if aws_managed and (not request.sandbox_provider or not provider_secret_name):
+        raise ValueError("Managed runs require a sandbox provider and provider secret name")
```
JarettForzano, line 86: Helper method to consolidate logic

services/tracker/src/tracker/utils/run_orchestration.py
```diff
@@ -243,47 +249,98 @@ def _parse_start_benchmark_request(payload: dict[str, Any]) -> StartBenchmarkReq
...
+        dry_run_lambda(runtime.clients, request.lambda_function)
+    return sandbox_provider_config
+
+
 # Pin the Taskiq task name to its pre-refactor value so in-flight messages
 # enqueued as `tracker.utils:process_benchmark` still match after the module move.
 @broker.task("tracker.utils:process_benchmark")
-@logfire.instrument("process_benchmark", extract_args=("benchmark_id_str", "verified_task_ids"))
+@logfire.instrument("process_benchmark", extract_args=False)
 async def process_benchmark(
```
JarettForzano, line 376: I feel like the code is hard to read since we are wrapping a lot of stuff. Can we use decorators or something to reduce wrappers?

JarettForzano (commented):
I kinda feel like we are coupling AWS too much with the code, it seems like we are wrapping everything which is making things hard to read and maintain. is there a way to reduce this sort of behavior? LMK if you think this is a non-problem, it just looks like we are making "managed" AWS a requirement for everything.

JarettForzano (approved):

PR #549, opened 2026-07-14
fix: restore cli aws credential chain

BradenBug:
## Summary

- use the AWS SDK credential chain for CLI S3 operations when Valkyrie has no explicit access-key pair
- preserve the existing explicit access-key path
- reject incomplete explicit credential configuration

## Why

Agent publishing jobs configure temporary AWS credentials in the process environment and keep only the S3 bucket and region in `valkyrie.yaml`. The CLI currently requires the access key and secret in that file, so it fails before boto can use the environment-backed OIDC credentials.

This keeps explicit access-key authentication available while allowing OIDC, SSO, profiles, and other SDK providers to supply and refresh credentials without persisting them in Valkyrie configuration.

## Stack impact

This PR targets `dev` below #532. After it merges, #532 should be restacked while retaining this credential selection behavior; #543, #544, and #545 should require only mechanical restacks.

## Verification

- `uv run pytest tests/unit/test_s3_config.py -q` — 4 passed
- `make format-check`
- `uv run ruff check .`
- `make typecheck`
- `make unit-test` — 205 passed, with the existing `test_machine_json_subprocess_suppresses_import_time_dotenv_warnings` failure reproducing identically on unmodified `origin/dev`


JarettForzano (approved):
Approved

PR #551, opened 2026-07-14
sdk: stream retained run evidence

tthuwng:
## Summary

- expose typed retained run metadata through the Valkyrie SDK
- expose run-output TARs as a context-managed async byte stream without buffering successful responses
- add public `RetryPolicy.ALLOW` / `RetryPolicy.FORBID` to run starts and persist it in Tracker run arguments/results
- reject manual retry/resume for a persisted `FORBID` run before task mutation
- atomically claim a `FORBID` task from `PENDING`, making duplicate or stale Taskiq deliveries read-only no-ops
- prevent terminal errors from being overwritten by a stale worker and reject queued-policy drift
- disable task-level sandbox setup retry, evaluator resume, and arbitrary Agent `install_cmd` replay for `FORBID` runs
- retain existing `ALLOW` behavior and idempotent bundle/bootstrap, desired-state egress, and read-only transport retries
- reserve SDK version 0.1.2 because PR #548 reserves 0.1.1

## API

```python
from valkyrie.sdk import RetryPolicy

started = await client.runs.start(
    agent,
    benchmark,
    retry_policy=RetryPolicy.FORBID,
)
metadata = await client.runs.metadata(started.benchmark_id)

async with client.runs.outputs(started.benchmark_id, task_ids=["001"]) as chunks:
    async for chunk in chunks:
        ...
```

Legacy callers and stored rows default to `allow`.

## No-retry invariants

- persisted Tracker policy is authoritative; a mismatched queued request cannot claim or mutate the task
- exactly one concurrent delivery can perform `PENDING -> BUILDING`
- `FORBID` duplicates in BUILDING, IN_PROGRESS, EVALUATING, ERROR, or FINISHED do not create an ErrorResult, EvaluationResult, sandbox, model call, or evaluator call
- `BUILDING -> IN_PROGRESS -> EVALUATING -> FINISHED` uses expected-state compare-and-swap guards for `FORBID`
- the first setup/install failure produces one terminal task error; no task/evaluator/install replay is attempted

## Verification

- full Tracker unit suite at the implementation checkpoint: `298 passed`
- post-audit affected Tracker suites: `90 passed`
- SDK public API and request serialization: `27 passed`
- concurrent duplicate-delivery, stale-state, queued-policy mismatch, ERROR transition, and one-shot install controls pass
- Ruff lint and format passed across SDK, Tracker, and tests
- BasedPyright: 0 errors and 0 warnings across changed production surfaces
- `git diff --check` passed
[...]

PR #552, opened 2026-07-14
test: diagnose Daytona egress transition latency

tthuwng:
## Diagnostic scope

Temporary, test-only branch stacked on #551. It does not change production code.

- Pre-resolves and confirms a stable off-list IPv4 endpoint while egress is open.
- Measures three-consecutive-round readiness for allowlisted DNS, cached off-list IP, and off-list DNS.
- Repeats five transitions with the current three-field clear and five with Daytona's documented `network_block_all=False` clear.
- Prints every observation and latency through a focused `-s` integration job.

The PR will be closed after the live evidence is captured. The purpose is to distinguish firewall propagation from DNS restoration/clear-payload behavior before choosing a production fix.

Local static gates: Ruff, format, strict BasedPyright, and diff check pass.

### DIFF
--- a/.github/workflows/tracker-integration-tests.yaml
+++ b/.github/workflows/tracker-integration-tests.yaml
@@ -50,7 +50,7 @@ jobs:
       - name: Run integration tests
         run: |
           cd services/tracker
-          uv run pytest tests/integration -n auto
+          uv run pytest tests/integration/test_sandbox.py::TestSandboxOperations::test_daytona_egress_transition_diagnostic -s
         env:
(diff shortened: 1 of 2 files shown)


tthuwng:
Diagnostic complete; this test-only PR is intentionally not for merge. Attempt 2 measured all five current clear cycles restoring direct-IP and DNS access within 0.1-2.0s, while restriction updates returned before the policy became active (first blocked observation around 2.5s). The one-field canonical clear left the prior allowlists installed and remained restricted past 30s. Evidence: https://github.com/vals-ai/Valkyrie/actions/runs/29333511411/attempts/2

PR #554, opened 2026-07-14
Rotate matching benchmark-service auth with hosted API key

conjfrnk:
## Why

Hosted tracker auth and benchmark-service auth are stored separately. Updating the hosted API key left per-benchmark credentials derived from the old key stale, so retries kept receiving 401 responses unless every mapping was edited manually.

## Change

Both `valk config set api_key ...` and hosted `valk config init` now rotate only `benchmark_auth` values that exactly match the previous key in raw or `Bearer ...` form. Independent per-service credentials are preserved, and output never displays credentials.

## Verification

1. Reproduced Unauthorized failures on exact Legal, EMB, and Terminal tasks launched before key rotation.
2. Confirmed current error cohorts and separated 401s from service-disconnect errors.
3. Traced both config entry points, `benchmark_service_headers`, and retry/resume header construction.
4. Confirmed exact tasks retried after explicit benchmark auth passed the former 401 point.
5. Added raw, Bearer, and independent-key regression coverage for both `config set` and hosted `config init`.

Checks run on the final diff:

- `uv run pytest tests/unit -q` — 209 passed
- `uv run basedpyright` — 0 errors
- full Ruff check and format check — passed
- `git diff --check` — passed

Current frozen tasks still need targeted retries after the credential is corrected; this change prevents future key rotations from leaving matching mappings stale.

### DIFF
--- a/src/valkyrie/cli/config/settings.py
+++ b/src/valkyrie/cli/config/settings.py
@@ -67,6 +67,7 @@ def init() -> None:
 
     if mode == "hosted":
         api_key = (os.environ.get("VALKYRIE_API_KEY") or click.prompt("API Key")).strip()
+        _rotate_matching_benchmark_auth(current_config, api_key)
         current_config["api_key"] = api_key
 
         # Validate the key and create/confirm org (uses default tracker URL)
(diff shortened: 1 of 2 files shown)


Chen-Oliver (commented):
apply the same change for `valk config init` too

Chen-Oliver (approved):

PR #558, opened 2026-07-15
feat: sandbox scheduler

som1shi:
## Summary

Integrates the PostgreSQL scheduler primitives from [Valkyrie #612](https://github.com/vals-ai/Valkyrie/pull/612) into Tracker runtime, CLI/SDK, deployment, and the scheduler overview. Existing `Task` rows remain the only queue state; queueing is feature-flagged, while unsupported or unmanaged providers keep direct execution.

## Changes

- Create queued `Task` rows before worker dispatch and schedule by provider pool, priority `0..4`, FIFO order, and run concurrency.
- Default queued runs to priority `3` and concurrency `5`.
- Keep existing execution, evaluation, retry, stop, finalization, and cleanup paths after admission.
- Preserve shared task state when a failed dispatch still has an active sibling; only the exact dispatch-owned `EVALUATING` attempt is failed.
- Fence evaluation recovery and ownership transfer with PostgreSQL advisory locks and exact attempt tokens.
- Add `SANDBOX_QUEUE_ENABLED`; shared Daytona capacity queues while unmanaged Daytona and Modal run directly.
- Add CLI, Tracker, and SDK priority support plus organization-scoped `GET /scheduler/overview`.
- Add the canonical Tracker OpenAPI generator, committed snapshot, drift test, and scheduler contract assertion.
- Pass queue configuration through Docker/deployment config and bump the SDK to `0.2.3`.

## Stack

1. [create-benchmark-service #128](https://github.com/vals-ai/create-benchmark-service/pull/128) → `main`
2. [Valkyrie #612](https://github.com/vals-ai/Valkyrie/pull/612) → `dev`
3. **[Valkyrie #558](https://github.com/vals-ai/Valkyrie/pull/558)** → currently `ss/scheduler-primitives`; retarget to `dev` after #612 merges (this PR)
4. [dashboards #4](https://github.com/vals-ai/dashboards/pull/4) → `main`

## Testing

- GitHub `tracker-unit-tests` on head `e7906bec7ebc7084c2503baca2b112efb2155263`: 675 passed at 88.76% coverage.
- Local PostgreSQL scheduler integration file on the same head: 17 passed.
- GitHub Tracker live integration, CLI tests/smoke, executor/package builds, Python 3.13/3.14 SDK compatibility, Ruff, BasedPyright, all lock validations, and Codecov patch coverage passed on the same head.
- Infrastructure checks are required to pass before merge. CDK synth is intentionally blocked while the PR targets `ss/scheduler-primitives`; retargeting to `dev` after #612 merges activates the required DEV synth and maintenance classification.
- [Live DEV E2E at `6bb777f`](https://github.com/vals-ai/Valkyrie/pull/558#issuecomment-5108079636): 80 runs across 10 clients, 8 services, 3 workers, and priorities P0–P4. All 326 queue snapshots preserved priority/FIFO and default P3; there were no duplicate claims, stuck rows, or leaked sandboxes. Fifty runs finished and 30 ended in explicit service or test-organization capacity errors.


src/valkyrie/cli/run/start.py
```diff
@@ -228,6 +241,13 @@ def resolve_webhook_config(
     default=None,
     help="Named sandbox provider from config (e.g., daytona, modal)",
 )
+@click.option(
+    "--priority",
+    type=click.IntRange(0, 4),
+    required=False,
+    default=None,
+    help="Opt into the queue at priority 0 (highest) through 4 (lowest); use 3 for normal priority",
```
BradenBug: "Opt into queue" - the queue should be enforced if enabled, no? Why would we allow people to not use the queue?

BradenBug (commented):
How is this configured? "Enable it with --use-queue; direct execution remains the default." - this is a CLI command?
So the queue is built / deployed regardless of deploy config and people can just opt out of it?

src/valkyrie/cli/run/start.py
```diff
@@ -180,9 +193,9 @@ def resolve_webhook_config(
 @click.option(
     "--concurrency",
     type=click.IntRange(min=1),
-    default=5,
+    default=None,
```
tthuwng: [P1] Preserve the legacy bounded default. This changes every omitted-concurrency run from 5 tasks to unlimited, including direct runs where `priority` is omitted. Large existing benchmarks can therefore fan out all task coroutines and eventually all sandboxes, changing cost and load independently of queue adoption. It also breaks new-client → old-Tracker starts because the new payload serializes `concurrency: null` into the old `int = 5` field; with an explicit concurrency, an old Tracker can instead ignore the unknown `priority` and silently execute directly. Please retain the default of 5 (or require an explicit unlimited setting) and add mixed-version request tests.

services/tracker/src/tracker/utils/run_orchestration.py
```diff
@@ -272,7 +273,22 @@ async def process_benchmark(
         org = session.exec(select(Org).where(Org.id == benchmark_row.org_id)).one()
 
     finalization_deferred = False
+    queue_context: DaytonaQueueContext | None = None
+    queue_redis: Redis | None = None
     try:
+        if start_benchmark_request.priority is not None:
+            organization_id = provider_secret.get("DAYTONA_ORGANIZATION_ID")
```
tthuwng: [P1] This introduces a required production integration that is absent from the provider documentation and rollout path. Queued runs now need both `DAYTONA_ORGANIZATION_ID` and an API key allowed to read organization usage; the current capacity classifier treats 401/403 as fatal. The existing provider guide still documents only API key, URL, target, and Sandboxes/Snapshots permissions. Please add the secret migration and usage-read permission to the provider/runbook docs and require an exact-head smoke with the production key before enabling queued runs.

services/tracker/src/tracker/utils/task_execution.py
```diff
@@ -586,15 +623,55 @@ def task_is_stopped() -> bool:
...
+                        return commit_task_status_transition(
+                            task_row.id,
+                            task_session,
+                            org,
+                            TaskStatus.BUILDING,
+                            expected_started_at=attempt_started_at,
+                            expected_status=TaskStatus.PENDING,
+                        )
+
+                sandbox = await enter_queued_sandbox(
```
tthuwng, line 1067: [P2] The stacked `enter_queued_sandbox` implementation awaits `create()` before invoking `mark_building`. That leaves the task `PENDING`—and visible as waiting—for the entire sandbox build, then makes `BUILDING` nearly instantaneous before `IN_PROGRESS`. Please transition to `BUILDING` after capacity admission but before awaiting sandbox creation so the lifecycle and dashboard reflect what is actually happening.

services/tracker/src/tracker/utils/task_execution.py
```diff
@@ -586,15 +623,55 @@ def task_is_stopped() -> bool:
...
+                    stack=sandbox_stack,
+                    context=queue_context,
+                    ticket=ticket,
+                    demand=normalize_task_resources(task_data.resources),
+                    create=sandbox_context,
+                    stopped=queued_task_ownership_lost,
```
tthuwng: [P2] This callback opens a synchronous SQLModel session on every queue poll. Combined with the one-second gate loop and the new unlimited default, a run with N waiting tasks produces roughly N PostgreSQL ownership reads per second plus Redis traffic. The live proof uses five one-task runs, so it does not exercise this boundary. Please keep waiter admission bounded or replace per-task database polling with a centralized/notification-based ownership check, and add a large-queue load test.

tthuwng (commented):
Comment-only review on exact head `642c3bf5590e1ebed065bbf48336b8901d2f1799`. These comments focus on preserving existing direct-run behavior, production provider integration, queue scalability, and lifecycle accuracy. I did not duplicate the existing unresolved thread about whether queue admission should be optional.

src/valkyrie/cli/run/start.py
som1shi: Agreed. I changed this so queueing is configuration-driven, not a per-run opt-in. A Daytona provider secret containing `DAYTONA_ORGANIZATION_ID` now forces queued admission for every new run using it; an omitted priority resolves to `3`, while `--priority 0..4` only overrides ordering. Provider secrets without the org ID remain on the direct path for personal/user-owned keys. The CLI help and output no longer say “Opt into queue.”

services/tracker/src/tracker/api/scheduler_overview.py
```diff
@@ -0,0 +1,267 @@
...
+    SchedulerOverviewResponse,
+    SchedulerPoolResponse,
+    SchedulerSummaryResponse,
+    SchedulerWaitingEntryResponse,
+)
+
+router = APIRouter(prefix="/scheduler")
+
+_ACTIVE_STATUSES = (TaskStatus.BUILDING, TaskStatus.IN_PROGRESS, TaskStatus.EVALUATING)
+_QUEUE_KEY = re.compile(r"^sandbox-queue:([0-9a-f]{24}):p([0-4])$")
```
OrestesK: **[P2] Can we keep the Redis queue protocol behind one owner?**

This module independently defines the queue-key regex, task-member decoding, heartbeat hashing, priority traversal, live-ticket filtering, and ordering already owned by `RedisQueueGate`. A storage change can therefore leave admission correct while `/scheduler/overview` silently diverges.

Verification: I compared `scheduler_overview.py:36-105` with `scheduler/gate.py:59-86,147-161` at exact head `f4a010ba`. This is structural ownership, so a runtime probe would not remove the duplication.

Please expose a narrow read-only ticket snapshot or shared protocol codec from the scheduler layer, leaving this endpoint responsible only for organization filtering and database enrichment.

OrestesK (commented):
Verified against exact head `f4a010ba`. One source-of-truth/maintenance finding; the existing Daytona rollout thread remains applicable and was not duplicated.

services/tracker/main.py
```diff
@@ -299,6 +304,30 @@ async def start_benchmark(
         }
     )
 
+    provider_config = await asyncio.to_thread(
```
OrestesK: Can we avoid resolving the provider when `SANDBOX_QUEUE_ENABLED` is false?

This unconditional secret fetch and provider construction changes the existing direct path: a queue-disabled start can now block or fail on AWS/provider credentials before the worker starts. The base endpoint had no provider-resolution dependency.

Please branch before this lookup:

- disabled with no priority: preserve the existing direct path;
- disabled with explicit priority: reject immediately;
- enabled: resolve the provider and inspect its admission capability.

OrestesK (commented):
Assuming priority is meant to apply across all pending tasks, the local limiter currently prevents that.

`TrackedTask.run()` holds the per-run limiter around the entire `process_task()` coroutine, but Redis admission happens inside that coroutine. With concurrency 5, only five tasks from a run are visible to Redis. Task 6 remains invisible while a later, lower-priority task from another run can join Redis and start first. `/scheduler/overview` also omits that hidden task.

We should either:

- register every prioritized task globally before local concurrency blocks it, then coordinate per-run slots separately; or
- explicitly define this as priority only among tasks already admitted by each run’s concurrency limit.

Simply moving the limiter is not enough because it currently bounds the task’s full execution lifecycle.

services/tracker/src/tracker/utils/task_execution.py
som1shi: Fixed in the PostgreSQL admission path. `claim_eligible_task` commits `PENDING -> BUILDING` before awaiting sandbox creation; `_start_claimed_task` moves it to `IN_PROGRESS` afterward. The scheduler integration test observes `PENDING`, `PENDING`, then `BUILDING` across the capacity checks and `create()`.

services/tracker/src/tracker/database/models.py
```diff
@@ -158,6 +158,8 @@ class BenchmarkArguments(BaseModel):
 
     contract: AgentContractRequest
     concurrency: int
+    priority: int | None = None
+    queue_pool_id: str | None = Field(default=None, exclude=True)
```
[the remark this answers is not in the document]
som1shi: Fixed. Tracker and SDK `BenchmarkArguments.priority` now require a strict integer in `0..4`, with focused boundary and invalid-value coverage.

services/tracker/src/tracker/utils/run_orchestration.py
```diff
@@ -52,6 +55,173 @@
...
+        session.commit()
+
+    await recover_queued_pool(queue_context)
+    monitor_task = asyncio.create_task(monitor.track_tasks())
+    cancellation_attempts: dict[UUID, datetime] | None = None
+
+    def launch(task_row: Task) -> None:
+        tracked_task = TrackedTask(
+            process_task(
+                task_row,
```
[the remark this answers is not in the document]
som1shi: Verified this path: `SandboxSetupError` is retried inside `_process_task_attempt`, and any ordinary exception that still escapes is caught by `TrackedTask.run`, which records the task as `ERROR` and returns `{task_id: None}`. It therefore does not escape `await runner` or cancel peer runners. No code change is needed for the described failure mode.

services/tracker/main.py
```diff
@@ -299,6 +307,43 @@ async def start_benchmark(
...
+        try:
+            provider_pool_id = provider.admission_pool_id
+        finally:
+            await provider.close()
+
+        # Admission identity is configured per provider secret; unmanaged credentials remain direct.
+        if provider_pool_id is not None:
+            if request.priority is None:
+                request = request.model_copy(update={"priority": 3})
+            resolved_queue_pool_id = queue_pool_id(provider_pool_id)
```
OrestesK, line 588: Could we reject targeted snapshots from queued admission for now? This pool ID is fixed from the provider's default Daytona target before task data is loaded, but `TargetedSnapshotSource` later creates via `source.target`; CBS's current admission check also returns `True` for snapshot sources. So a cross-target snapshot is queued under the wrong pool and isn't capacity-checked for its real target. Full support needs source-aware, per-task pool identity, so an explicit unsupported error seems safer for this PR.

som1shi: Queued admission now rejects targeted snapshots before checking capacity or claiming a provider slot. Supporting them later will require source-aware pool identity.

services/tracker/src/tracker/utils/task_execution.py
```diff
@@ -652,30 +700,57 @@ def task_is_stopped() -> bool:
...
+                volumes=task_data.volumes,
+                creation_semaphore=creation_semaphore,
+                unique_name=queue_context is None,
+            )
+
+        async with AsyncExitStack() as sandbox_stack:
+            if queue_context is None:
+                sandbox = await sandbox_stack.enter_async_context(sandbox_context())
+            else:
+                sandbox = await enter_queued_sandbox(
```
OrestesK, line 1067: Can we carry `ExecutionAuthority` through queued admission?

The direct path fences `PENDING → BUILDING → IN_PROGRESS` against the current dispatch, but `enter_queued_sandbox()` only checks the task attempt and benchmark status. After Retry, Resume, or force Stop revokes a dispatch, the old executor can still claim the queue slot and create a sandbox until TaskMonitor or ExecutorHost notices asynchronously.

Please validate the exact dispatch authority in the same transactions as the queued claim and start transitions, and add a test where authority is revoked during admission.

services/tracker/src/tracker/utils/run_orchestration.py
```diff
@@ -51,6 +55,177 @@
...
+            .where(Task.status == TaskStatus.EVALUATING)
+            .with_for_update()
+        ).all()
+        for task_row in resumable_evaluations:
+            if task_row.eval_resume_state is None:
+                continue
+            resumed_at = datetime.now(task_row.started_at.tzinfo)
+            if resumed_at <= task_row.started_at:
+                resumed_at = task_row.started_at + timedelta(microseconds=1)
+            task_row.started_at = resumed_at
```
OrestesK: Can we avoid advancing `started_at` after the owning dispatch has already been created?

ExecutorHost failure cleanup treats `Task.started_at <= ExecutorDispatch.created_at` as the ownership boundary. This update makes the resumed attempt newer than its dispatch, so a later executor failure can skip a task that the dispatch actually owns.

The attempt should either be rebased before recovery dispatch admission, or task attempts should record their owning dispatch explicitly. Please cover executor failure after this resume setup.

services/tracker/main.py
```diff
@@ -916,16 +967,45 @@ async def retry_or_resume_benchmark(
...
+            if any(
+                task_row.status == TaskStatus.IN_PROGRESS
+                or (task_row.status == TaskStatus.EVALUATING and task_row.eval_resume_state is None)
+                for task_row in scheduler_rows
+            ):
+                raise HTTPException(
+                    status_code=409,
+                    detail="Run has active tasks that cannot be resumed safely; stop the run before resuming",
+                )
+            recovery_task_ids = [task_row.task_id for task_row in scheduler_rows]
```
OrestesK, line 1295: Can we atomically transfer ownership of resumable evaluations before admitting this recovery dispatch?

For an in-progress queued run, this path accepts an `EVALUATING` task with durable resume state while `admit_recovery_dispatch()` deliberately keeps the original dispatch active. Both executors can therefore reach `resume_evaluation()` for the same task before the later timestamp/status CAS cancels one. That CAS protects database completion, but not duplicate external evaluation work.

Please claim the evaluation attempt for one dispatch before the external call, or explicitly supersede the old owner, and test two live coordinators trying to resume the same evaluation.

OrestesK (commented):
The strict global queue needs an explicit hard-loss policy before this is enabled. A hard ExecutorHost/process loss after a dispatch is claimed `RUNNING`, but before its task leaves `PENDING`, leaves that task as the global queue head indefinitely. Other benchmarks in the provider pool cannot pass it. Redis redelivery does not repair this because dispatch claim only accepts `QUEUED`; the redelivered `RUNNING` dispatch becomes a successful no-op.

#601 intentionally excluded automatic orphan repair, which was a per-run limitation there. With global FIFO it becomes a provider-pool outage, so this needs a deliberate choice: automatic replacement, adoptable work, a bounded bypass rule, or accepted manual recovery with alerting. Whichever policy we choose, can we add a failure-injection test with an unrelated same-pool contender behind the lost dispatch?

Can we also replace the old live-test evidence with an exact-head ExecutorHost/PEX run? The recorded 80-run E2E predates immutable Executor artifacts, the final CBS admission dependency, and release-based routing. The current `executor-build` check stops at lock validation, so it has not built or checked the actual PEX. After fixing the lock, the deployment proof should cover queue-enabled execution through the PEX, admission success/failure, targeted-snapshot rejection, release promotion with an older queued dispatch on the draining release, and authority revocation or hard process loss.

The SDK priority change also needs a version bump: the current package check fails because this branch still declares `0.2.3`, equal to the base version rather than greater than it. The manual-on-PR live-test failure and maintenance-classification failure are expected.

services/tracker/src/tracker/utils/task_execution.py
som1shi: Implemented. Queued admission now receives `ExecutionAuthority` and validates the exact RUNNING dispatch in the same database transaction as both `PENDING → BUILDING` and `BUILDING → IN_PROGRESS`. The claim/start helpers no longer commit internally. If authority is revoked after sandbox creation, admission closes the sandbox with cancellation shielding and never starts the task. The real-PostgreSQL `test_revocation_before_claim_creates_no_sandbox` and `test_revocation_during_creation_cleans_up_without_starting` cases cover both boundaries.

services/tracker/src/tracker/utils/run_orchestration.py
som1shi: Fixed by removing the post-admission clock bump. Recovery now transfers every resumable `EVALUATING` task to the new dispatch in the same transaction with `Task.started_at == ExecutorDispatch.created_at`, and rejects a partial transfer. Dispatch failure requires that exact token for `EVALUATING` attempts while retaining the existing upper-bound rule for pre-evaluation states. The recovery-token and dispatch-failure tests cover both ownership paths.

services/tracker/main.py
som1shi: Implemented with one shared task-evaluation advisory-lock identity. Recovery row-locks candidates and nonblockingly probes `pg_try_advisory_xact_lock` before any dispatch/task mutation; an active owner returns `409` with no mutation. The admitted dispatch and exact `started_at` token are committed together. The executor then holds the matching session advisory lock across `resume_evaluation()`, resume-state callbacks, and terminal/error persistence, rechecking dispatch authority and the exact token before the external call. Real-PostgreSQL coverage proves held-lock `409`, two recovery handoffs leave one external caller, stale coordinators make no external call, callback/finalization work with a one-connection pool, and cancellation releases the lock. This retains the approved bounded connection-loss risk rather than adding an external protocol.

som1shi:
@OrestesK Final stacked-head re-review requested on `e7906bec7ebc7084c2503baca2b112efb2155263` against #612 head `499655300c268df7bd8182d3357d5484573de766`. The sibling-safe dispatch recovery fix, canonical OpenAPI contract, current-dev synchronization, and merged scheduler recovery tests are included. Tracker unit CI is green (675 passed, 88.76%); all functional/static/lock checks are green or completing. CDK synth remains intentionally blocked only because this stacked PR still targets `ss/scheduler-primitives`; it will be retargeted to `dev` after #612 merges.

services/tracker/generate_openapi.py
```diff
@@ -0,0 +1,72 @@
...
+
+from copy import deepcopy
+import json
+from pathlib import Path
+from typing import Any
+
+from main import app
+
+OPENAPI_PATH = Path(__file__).with_name("openapi.json")
+BEARER_OR_API_KEY: list[dict[str, list[str]]] = [{"BearerAuth": []}, {"ApiKeyAuth": []}]
```
OrestesK, line 11: why do we need openapi?

services/tracker/src/tracker/sandbox.py
```diff
@@ -259,6 +259,8 @@ async def create_sandbox(
     env_vars: dict[str, str] | None = None,
     volumes: list[VolumeMount] | None = None,
     sandbox_secrets: dict[str, str] | None = None,
+    *,
+    unique_name: bool = True,
```
OrestesK, line 257: why needed

services/tracker/generate_openapi.py
som1shi: I use this to bridge the dashboard and scheduler and keep their types synchronized across repositories.

services/tracker/src/tracker/sandbox.py
som1shi: This lets retries reuse the same sandbox instead of creating a duplicate with a new random name.

OrestesK (approved):

PR #570, opened 2026-07-16
change: deploy valkyrie dev in its own account

BradenBug:
## Why

Valkyrie dev currently runs in the same AWS account as production, so stage names do not provide an account boundary. The existing four dev stacks need to deploy into a dedicated account in `us-east-1` without changing production resources.

## What changed

- Keep production push deployment intact. Run dev operations only through a protected manual workflow from `dev`, with an account allowlist and STS identity check before checkout or synthesis.
- Reuse the same account and Region guard for local `plan` and `deploy` Make targets.
- Make the existing Shared and Tracker stacks account-portable. Dev imports its operator-created child hosted zone and certificate, uses an account-qualified hardened bucket, and keeps RDS non-public.
- Read the dev Descope project ID from the protected GitHub `dev` Environment and require it before planning or deploying. Keep the matching management key in account-local Secrets Manager.
- Leave Sentry disabled in dev unless the protected Environment names an account-local DSN secret. Production keeps its existing Sentry configuration.
- Keep account bootstrap, GitHub OIDC, DNS, and certificate ownership outside the application stack graph. The application still has only Shared, Tracker, Worker, and Monitoring stacks.
- Publish account-local Shared and Tracker discovery parameters for the separately reviewed benchmark-service integration.
- Use the standard GitHub checkout because this repository has no submodules or private checkout dependency.

## How to review

Start with `.github/workflows/deploy.yaml` and `infra/deployment_target.py` for the authorization path. Then review the dev-only branches in `infra/shared.py` and `infra/tracker_stack.py`.

The production synth comparison is the primary guard for the stage branches: all four production templates are unchanged.

## Verification

- `cd infra && make test` — 18 tests passed.
- `cd infra && make lint` — 14 files passed the format check; Ruff checks passed.
- `cd infra && make typecheck` — 0 errors, 0 warnings, 0 notes.
- The deploy workflow parsed with `yq`; its changed validation shell passed `shellcheck`.
- A dev synth with a synthetic account and `DESCOPE_PROJECT_ID` produced Shared, Tracker, Worker, and Monitoring successfully without a parent-zone lookup.
- Production synths before and after the change used the same account, Region, and context. `difft` reported no changes for all four templates.

No live application deployment was run. The account-local child zone, certificate, Descope management key, and protected Environment project ID must exist before the first dev deployment.


infra/deployment_target.py
```diff
@@ -0,0 +1,127 @@
...
+import re
+from collections.abc import Mapping
+from dataclasses import dataclass
+from typing import Protocol, cast
+
+from boto3.session import Session
+from botocore.exceptions import BotoCoreError, ClientError
+from mypy_boto3_sts import STSClient as BotoStsClient
+
+PRODUCTION_ACCOUNT_ID = "613431292675"
```
BradenBug: why is the production account ID hardcoded?

BradenBug: [from Codex]: Moved the Vals production account ID out of reusable Python code and into deployment configuration in 01f82d10. The validator now reads PRODUCTION_ACCOUNT_ID; the Makefile supplies an overrideable Vals default, and the dev workflow supplies its hosted value. The dev-versus-production safety check remains.

JarettForzano (approved):

PR #578, opened 2026-07-17
Add canonical run naming with legacy compatibility

som1shi:
## Summary

- add canonical `/runs` routes, `run_id` fields, and `Run*` interfaces across the CLI, Tracker, and SDK
- keep legacy HTTP routes, request fields, response shapes, Lambda keys, fallback behavior, Taskiq registration, and existing telemetry names at external compatibility boundaries
- keep fallbacks limited to exact unmatched-route 404 responses so authentication, validation, and resource-not-found errors still surface
- keep benchmark-definition/service terminology and physical database, Alembic, S3, CloudWatch, queue, and stored-artifact names unchanged
- keep the newly added SDK canonical under `client.runs`, add `client.runs.update`, and publish the changed package as `0.2.2`
- remove legacy inheritance from canonical Tracker request models and explicitly translate only at legacy handler boundaries
- finish internal run naming with `RunContext`, `process_run`, `start_request_to_run_row`, `fetch_filtered_run_rows`, `build_run_table_rows`, and related shared helpers without retaining internal-only aliases
- update Slack notifications to say Run Update/Complete/Error/Stopped while retaining the benchmark definition name
- carry run-level errors through canonical Tracker/SDK/CLI responses, dual-emit `run_id` plus legacy `benchmark_id` telemetry, and prevent stale all-error finalizers from overwriting successful retries

## Verification

- CLI unit and local integration: 208 passed; coverage gate passed (84.73%)
- Tracker unit, local integration, and Alembic: 413 passed; coverage gate passed (87.11%)
- SDK and Tracker contract: 160 passed
- root Ruff, format check, and configured basedpyright: clean
- targeted strict Tracker check: no new diagnostics; the existing `retrieve_results` diagnostics remain non-gated
- SDK version gate: `0.2.1` base to `0.2.2` current
- SDK wheel and sdist: built; metadata, Twine, wheel contents, isolated installs, packaged tests, namespace coexistence, and examples verified
- full compatibility-boundary audit and final full-range code review

## Release coordination

`dev` contains SDK `0.2.1`, so this PR uses `0.2.2`. Other open SDK PRs should refresh from `dev` before release.

`vals-ai/platform` still consumes the legacy Tracker routes and `benchmark_id` response fields. Those aliases must remain available through the consumer migration window.

Tracks vals-ai/benchmarks#2082.


som1shi:
Audit follow-up on head `836b9882`:

- Fixed the net-new Low finding in `_stream_with_legacy_fallback`: when no legacy-specific override is supplied, the legacy retry now reuses the canonical query params, matching the GET/POST fallback helpers.
- Added a regression test that failed on the previous head and verifies both canonical and legacy stream requests receive the shared params.
- The reported 84 strict-basedpyright errors under `services/tracker` predate this PR and are outside this rename migration; no suppression or unrelated type cleanup was added here.

Fresh verification: coverage-gated CLI/local-integration suite **193 passed** at **83.68%** coverage; root basedpyright and Ruff are clean.

PR #593, opened 2026-07-19
Improve all-task-error run messages

JarettForzano:
## Summary

Improve terminal runs where every task errors by deriving a useful run-level message with one representative for each distinct group of current task errors. Terminal fetch responses expose that stored message, while in-progress fetches continue returning no run-level error.

image

## Changes

- Use each task's latest error from the current run during all-task-error finalization.
- Normalize task-specific IDs and group similar messages with the standard-library `difflib.get_close_matches`.
- Store frequency-ordered representative errors for every distinct group, including one `N/N` entry when all task errors match.
- Return the stored terminal error through tracker, SDK, CLI text, and JSON/JSONL fetch responses.
- Share terminal-control escaping between error and fetch displays.

## Related Issues

N/A

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor
- [ ] Docs / config

## Testing

- [x] Manually tested

## Checklist

- [x] Self-reviewed the diff
- [x] Docs updated if needed


services/tracker/src/tracker/utils/run_orchestration.py
```diff
@@ -191,6 +192,38 @@ def fetch_final_score_inputs(session: Session, benchmark_row: Benchmark, org: Or
...
+    with Session(bind=engine) as session:
+        benchmark_row = fetch_benchmark_row(benchmark_id, session, org, for_update=True)
+        if has_runnable_tasks(session, benchmark_row, org):
+            return True
+        if has_stopped_tasks(session, benchmark_row, org):
+            set_benchmark_final_status(benchmark_row, session, org)
+            return False
+
+        # Mark the run as errored so future fetches return the discovered task errors.
+        commit_benchmark_error(benchmark_row, session, error_message)
```
som1shi, line 223: Could this overwrite a retry that finishes while the error summary is being built? Not sure but something to double check

JarettForzano: technically there exists this code here

```python
if has_runnable_tasks(session, benchmark_row, org):
            return True
```

Which exists from a prev pr that introduced the race condition. its supposed to stop the rest of the code from executing if we have another benchmark process running.

I am going to rewrite our queue system at some point since its outdated. We technically want to be using SQS FIFO

som1shi: okay, shouldnt really be a problem then

som1shi (approved):

PR #601, opened 2026-07-20
feat: stable deploys

OrestesK:
## What changed

This PR replaces the legacy Worker execution path with stable ExecutorHost services that run immutable, digest-verified Executor artifacts. Normal Executor release activation can move new work forward without changing or interrupting work already in progress. PostgreSQL remains authoritative; Redis is delivery transport, and every dispatch records the exact bytes selected for execution.

- Adds the `CANDIDATE` → `ACTIVE` → `DRAINING` → `RETIRED` lifecycle. Exactly one release admits new work; any number of older releases may drain concurrently.
- Stores immutable initial release provenance and mutable current execution ownership on each benchmark, plus an immutable release/artifact/operation snapshot on each dispatch.
- Makes ExecutorHost claim each dispatch before downloading, verifying, caching, and running its artifact. Duplicate or non-claimable deliveries cannot start a second process.
- Adds API/SDK release identity, sealed deployment activation, automatic retirement, release-test/Driver infrastructure, scoped IAM, forward-only migrations, documentation, and diagrams. It removes the manual release lifecycle CLI, status command, and executor rollback workflow.

### Runtime contract

Release routing changes only at a whole-run terminal boundary. It does not change which tasks the existing retry, resume, and stop logic selects.

| Operation / state | Release used |
| --- | --- |
| New benchmark `START` | Current `ACTIVE` release |
| Original or retried work while `IN_PROGRESS` | Benchmark's current execution release, including while `DRAINING` |
| Partial stop while still `IN_PROGRESS` | No release handoff |
| Retry or resume while `STOPPING` | Rejected |
| Retry or resume from `STOPPED`, `FINISHED`, or `ERROR` | Current `ACTIVE` release, which becomes the new current execution release |

Promotion never rewrites existing ownership or queued/running dispatch snapshots. A draining release accepts no new starts but continues work it already owns. Tracker automatically retires each draining release after its queued/running dispatches, active owners, and unknown-owner blockers clear.

Activation is forward-only: after a release becomes active, recovery is another deployment of a new release. Normal ECS deployment circuit-breaker and CloudFormation rollback behavior remains unchanged. Artifact deletion is separate from metadata retirement: it requires a release to be retired for 30 days with no work blockers, and this PR adds no automatic cleanup job.

## Rollout

> [!WARNING]
> Normal Executor release activation does not interrupt work already in progress. The first legacy Worker → ExecutorHost cutover and changes classified by the required `maintenance-classification` CI check as Executor or database maintenance are different: they stop affected work. The CI check intentionally fails for maintenance changes; an authorized force merge approves the outage. During maintenance, admission closes and Start, Retry, Resume, and concurrency changes return `503` until deployment completes. Follow [`docs/RELEASES.md#first-executor-dispatch-cutover`](docs/RELEASES.md#first-executor-dispatch-cutover) for the one-time cutover. After the cutover, no legacy Worker or `taskiq` compatibility path remains.

- Dev Executor work runs automatically only when required by the change; core-only changes do not build or activate an Executor release. Production Executor work waits for approval on the protected `production-release` GitHub Environment.
- Plan, partial-scope, and credentials-only operations never publish or activate an Executor release.
- Migration `e9f0a1b2c3d4` is forward-only because current execution ownership is required runtime state. Do not downgrade across it or deploy a Tracker image that predates it after migration.
- Release-test uses dev-sized, stage-scoped resources. Its target guard also permits an explicit authorized campaign in the production account; unrelated account resources remain outside the release-test boundary. Secret values never travel in task overrides.

## How tested

Final staged-tree verification:

- `(cd infra && CDK_DEFAULT_ACCOUNT=123456789012 CDK_DEFAULT_REGION=us-east-1 PYTHONPATH=. uv run python -m unittest discover -s tests -p "test_*.py")` — **70 passed**.
- Focused runtime suites — **122 passed**.
- Executor artifact export, build, archive-validation, and runtime PEX smoke — passed.
- `actionlint`, Ruff, basedpyright, YAML parsing, lock, and diff checks — passed.
- Docker-backed PostgreSQL migration-lock integration — **not run locally because Docker was unavailable**.

Historical live architecture evidence:

- Isolated-stage campaign `20260725T233000Z` at source `74ff113c` passed **41/41 cases** and **126/126 assertions** across release/dispatch authority, pinned work, recovery, duplicate delivery, failure boundaries, and scoped infrastructure authority.
- That campaign predates the final deployment-remediation, CLI-removal, and automatic-retirement amendments. It supports the underlying architecture but is not final-diff proof; current local verification and its limitation are listed above.
- The campaign changed release-test stage resources only; it did not deploy this behavior to production services.

## Reviewer guide

Recommended review order:

1. Runtime and rollout contract: `docs/RELEASES.md` and `valkyrie-release-lifecycle.png`.
2. Release lifecycle: `services/tracker/src/tracker/release_control.py`, `release_entrypoint.py`, and `release_retirement.py`.
3. Admission and recovery: `services/tracker/src/tracker/dispatch_control.py`, `main.py`, `utils/run_orchestration.py`, and `utils/run_control.py`.
4. Persistence: release/dispatch models and migrations `c7d8e9f0a1b2`, `d8e9f0a1b2c3`, and `e9f0a1b2c3d4`.
5. Execution boundary: `services/executor_host/supervisor.py`.
6. Deployment boundary: `.github/workflows/deploy.yaml`, `infra/executor_release.py`, `tracker_stack.py`, `executor_stack.py`, and `driver_stack.py`.
7. Proof: release-control, retirement, dispatch-control, ExecutorHost, migration, and infrastructure tests.

## Explicit non-goals

This PR does not add or change:

- a global concurrency limit across releases;
- retry, resume, or stop task-selection rules;
- exactly-once delivery guarantees;
- automatic orphan recovery, replay, requeue, or repair;
- production rollout;
- queue-based autoscaling or executor cache optimizations.

## Design and documentation

- Canonical implemented contract and cutover runbook: `docs/RELEASES.md`
- [Deployment & Release Design — Notion](https://app.notion.com/p/3a38a877dd8d802bae9ff14730826e05)
- Architecture diagrams:
  - `valkyrie-release-coexistence.excalidraw`
  - `valkyrie-dispatch-ownership.excalidraw`
  - `valkyrie-release-lifecycle.excalidraw`

---

<a href="https://app.devin.ai/review/vals-ai/valkyrie/pull/601" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


services/executor_host/supervisor.py
```diff
@@ -0,0 +1,421 @@
...
+            user=self.user,
+            password=self.password,
+        )
+
+    async def claim(self, dispatch_id: str, benchmark_id: str, dispatch: ArtifactDispatch) -> bool:
+        return await asyncio.to_thread(self._claim, dispatch_id, benchmark_id, dispatch)
+
+    def _claim(self, dispatch_id: str, benchmark_id: str, dispatch: ArtifactDispatch) -> bool:
+        with self._connect() as connection, connection.cursor() as cursor:
+            cursor.execute(
```
BradenBug, line 236: **Blocker:** This claim fences only on the dispatch row, not on the benchmark's current execution. After terminal recovery advances `current_execution_release_id`, an older queued dispatch can still claim and run. `process_benchmark` then operates on whichever tasks are currently `PENDING`, and late cleanup from the old executor can mark the newer attempt `ERROR`. Please add a benchmark-level `current_dispatch_id` or execution epoch, advance it atomically during recovery, and require the claim plus every finalization and cleanup write to match it. Superseded queued dispatches should be cancelled.

services/tracker/src/tracker/database/migrations/versions/c7d8e9f0a1b2_add_executor_release_ownership.py
```diff
@@ -0,0 +1,94 @@
...
+        sa.PrimaryKeyConstraint("id"),
+    )
+    op.bulk_insert(
+        sa.table(
+            "executoradmission",
+            sa.column("id", sa.Integer()),
+            sa.column("release_id", sa.String()),
+            sa.column("updated_at", sa.DateTime()),
+        ),
+        [{"id": 1, "release_id": None, "updated_at": datetime.now(UTC)}],
```
BradenBug, line 59: This seeds admission with no active release, while every new start immediately returns 503 unless a release has been promoted. A merge to `dev` automatically deploys all stacks, but no workflow builds and uploads an executor artifact or runs register -> verify -> promote before traffic moves. Tracker health still passes, so ECS can complete the cutover while starts are unavailable. Please make artifact publication and verified activation a gated rollout step before switching producer traffic.

services/tracker/src/tracker/database/migrations/versions/e9f0a1b2c3d4_add_current_execution_release.py
```diff
@@ -0,0 +1,44 @@
...
+    op.create_foreign_key(
+        "fk_benchmark_current_execution_release_id_executorrelease",
+        "benchmark",
+        "executorrelease",
+        ["current_execution_release_id"],
+        ["id"],
+    )
+
+
+def downgrade() -> None:
```
BradenBug, line 40: **Blocker:** This forward-only revision is incompatible with the Tracker service's automatic ECS rollback. Every Tracker task runs `alembic upgrade head` on startup, while the deployment circuit breaker restores the previous image on failure. Once a new task advances the database, a relaunched pre-PR image cannot resolve this revision, so rollback is not a valid recovery path. Please separate migration from service rollout and use an old/new-compatible bridge or an explicit roll-forward-only procedure.

(I think we don't really support rollbacks as it is, we should look into that as well)

results-d01172e1-f598-4992-9477-c31730df129d.json
```diff
@@ -0,0 +1,61 @@
...
+            "run_cmd": "/usr/local/bin/python3 /bundle/ok-pi-agent/run_agent.py --problem-statement '{problem_statement_path}' --task-id '{task_id}' --timeout-seconds 1800",
+            "final_output": "/logs/ok-pi-agent",
+            "output_artifacts": [],
+            "egress_allowlist": []
+        },
+        "concurrency": 1,
+        "task_ids": [
+            "vals-ai-valkyrie-7ad2042d37a9"
+        ],
+        "dataset": "s3://agentic-harness/benchmarks/valsmith-datasets/vals-ai-valkyrie/665f2b07-3586-45e7-ae18-c869b676952e/manifest.json",
```
BradenBug: You'll delete all these artifacts before merging yeah?

services/tracker/main.py
```diff
@@ -356,14 +408,15 @@ async def start_benchmark(
...
 
-    await (
-        process_benchmark.kicker()
-        .with_labels(**_taskiq_labels())
-        .kiq(
-            start_benchmark_request_json=request.model_dump(),
-            benchmark_id_str=str(benchmark_row.id),
-            verified_task_ids=verify_response.task_ids,
-        )
+    executor_dispatch = create_executor_dispatch(benchmark_row.id, active_release, ExecutorDispatchKind.START)
```
BradenBug: This is not a durable outbox boundary. Start commits the active benchmark before this dispatch exists, then this transaction commits `QUEUED` before the Redis publication. A crash in either gap permanently strands the run: there is no scanner that republishes queued rows, and retrying an already-`IN_PROGRESS` benchmark returns without enqueueing again. Please commit benchmark ownership, immutable invocation data, and the dispatch atomically, then publish dispatch IDs through a retryable outbox relay.

services/executor_host/supervisor.py
```diff
@@ -0,0 +1,421 @@
...
+    executor_supervisor: ExecutorSupervisor,
+    store: ExecutorDispatchStore,
+    *,
+    executor_dispatch_id: str,
+    dispatch: ArtifactDispatch,
+    start_benchmark_request_json: Mapping[str, object],
+    benchmark_id_str: str,
+    verified_task_ids: list[str],
+) -> None:
+    claimed = await store.claim(executor_dispatch_id, benchmark_id_str, dispatch)
```
BradenBug: The claim is cancellation-unsafe. `asyncio.to_thread` can continue and commit `QUEUED -> RUNNING` after this coroutine has been cancelled, and cancellation can occur before entering the cleanup `try`. That leaves a `RUNNING` dispatch with no child and no terminalization; there is intentionally no reaper to recover it. Task protection is also acquired only after ownership and its failures are swallowed. Please make claim completion cancellation-safe, require protection before ownership, and add a lease or repair path for abandoned `RUNNING` rows.

services/executor_host/supervisor.py
```diff
@@ -0,0 +1,421 @@
...
+        await executor_supervisor.run(
+            dispatch,
+            start_benchmark_request_json=start_benchmark_request_json,
+            benchmark_id_str=benchmark_id_str,
+            verified_task_ids=verified_task_ids,
+            executor_dispatch_id=executor_dispatch_id,
+        )
+    except BaseException:
+        try:
+            await store.finish(executor_dispatch_id, succeeded=False)
```
BradenBug: Artifact download, digest, cache, or process-spawn failures terminalize only the dispatch. The benchmark remains `IN_PROGRESS`, potentially with zero task rows, while redelivery cannot reclaim the now-`FAILED` dispatch and release retirement continues treating the benchmark as active. Please either retain an explicitly retryable dispatch state or atomically move the benchmark to a recoverable error state, guarded by the current execution fence.

tests/unit/executor_host/test_supervisor.py
```diff
@@ -0,0 +1,364 @@
...
+from services.executor_host.supervisor import (  # pyright: ignore[reportMissingImports]
+    ArtifactDispatch,
+    ExecutorSupervisor,
+    parse_s3_uri,
+    run_executor_dispatch,
+    verify_file_digest,
+)
+
+
+class FakeDispatchStore:
```
BradenBug, line 28: These ExecutorHost tests do not currently gate the PR. The root `Makefile` test target runs only CLI and local integration suites, no workflow references `services/executor_host` or this test directory, and CI never builds the new Docker image. All current checks can therefore pass without exercising this service. Please add a service-owned Make target and workflow that runs these tests and builds the actual ARM64 image from the committed lock and requirements.

services/tracker/tests/integration/local/database/test_release_control.py
```diff
@@ -0,0 +1,433 @@
...
+        arguments=BenchmarkArguments(
+            contract=AgentContractRequest(name="race-agent", install_cmd="true", run_cmd="true"),
+            concurrency=1,
+        ),
+    )
+    pin_benchmark_to_release(benchmark_a, release_a)
+    postgres_session.add(benchmark_a)
+    postgres_session.commit()
+
+    def recover(session: Session, benchmark_id: UUID) -> None:
```
BradenBug, line 284: This race helper proves a different transaction from production. It selects and locks admission before locking the benchmark; `retry_or_resume_benchmark` locks the benchmark first and only later selects the active release. Start, retry, and stop are similarly reimplemented in this test file, so these tests can remain green when the route loses a lock or changes ordering. Please extract canonical orchestration callables used by the routes and drive those exact operations against PostgreSQL with deterministic barriers.

infra/worker_stack.py
```diff
@@ -175,6 +198,67 @@ def __init__(
             assign_public_ip=True,
         )
 
+        # ── Stable executor host ─────────────────────────────────────────
```
BradenBug: This becomes the primary execution path, but `MonitoringStack` still receives only the legacy worker service. ExecutorHost can be at zero tasks, disconnected, or accumulating queued and stale-running dispatches while every infrastructure alarm remains green. Please add first-class ExecutorHost service alarms and dashboards, stable-queue lag, stale dispatch-age signals, and explicit structured logging before switching producer traffic.

docs/RELEASES.md
```diff
@@ -0,0 +1,297 @@
...
+
+The Package R driver is a static Fargate task definition, not a service. It has a
+no-ingress security group, explicit VPC/database/Redis/DNS/HTTPS egress, retained
+logs, named secret references, and separate execution, task, and operator roles.
+The operator role can run only that task definition and pass only its two roles.
+Public IP assignment is a launch-time requirement because the stage has public
+subnets and no NAT gateway; it does not expose Tracker, whose ALB remains
+internal.
+
+Before synthesis, set:
```
BradenBug: This runbook cannot synthesize the release-test stack as written: `infra/driver_stack.py` also requires `RELEASE_TEST_SANDBOX_PROVIDER_SECRET_ARN`, but it is missing here. The infrastructure README also omits the new `driver` scope from its accepted-scope list. Please document the complete prerequisite, why it is required, the console alternative where one exists, and a concrete completion check for each operator step.

docs/executor-releases/README.md
```diff
@@ -0,0 +1,297 @@
+# Executor releases
```
BradenBug, line 1: This doc assumes the reader knows what an "executor release" is without defining or introducing the concept anywhere else in the documentation. This is a new construct in the service and should be at least mentioned in the Tracker / Infra documentation. It might be worth having a general DEPLOYMENTS.md doc instead of this.

docs/RELEASES.md
```diff
@@ -0,0 +1,297 @@
...
+Successive promotions are independent: A, B, and C may all drain concurrently,
+and each retires when its own active execution count reaches zero. There is no
+two-release limit.
+
+Do not delete a release artifact manually while `tracker.release_cli status`
+reports an active-execution or retention blocker.
+
+## Release-test
+
+The current release-test stage is dev-sized but runs in the production AWS
```
BradenBug: Why does it run in production? What is this?

services/tracker/src/tracker/release_cli.py
```diff
@@ -0,0 +1,73 @@
...
+
+    args = parser.parse_args(argv)
+    try:
+        with Session(engine) as session:
+            if args.command == "register":
+                release = register_release(
+                    session,
+                    ExecutorRelease(
+                        id=args.release_id,
+                        artifact_uri=args.artifact_uri,
```
BradenBug: Could we provision a dedicated, stage-scoped executor-release bucket in CDK and store only an object key here (or reject URIs outside that bucket)? This currently accepts an arbitrary S3 URI, but I cannot find an artifact publishing/upload path, while Tracker and ExecutorHost are granted read access only to the general `AgenticHarnessBucket`. These are executable artifacts retained for rollback, so their writer permissions, retention, and deletion boundary should be independent of unrelated benchmark artifacts. The release publisher should be the only writer; Tracker and ExecutorHost can remain read-only.

docs/executor-releases/README.md
```diff
@@ -0,0 +1,297 @@
...
+reports an active-execution or retention blocker.
+
+## Release-test
+
+The current release-test stage is dev-sized but runs in the production AWS
+account. Every operation must specify `STAGE=release-test`, account
+`613431292675`, and region `us-east-1`; unrelated account resources remain
+outside the release-test boundary.
+
+The Package R driver is a static Fargate task definition, not a service. It has a
```
BradenBug, line 321: What is Package R driver

docs/executor-releases/README.md
```diff
@@ -0,0 +1,297 @@
...
+logs, named secret references, and separate execution, task, and operator roles.
+The operator role can run only that task definition and pass only its two roles.
+Public IP assignment is a launch-time requirement because the stage has public
+subnets and no NAT gateway; it does not expose Tracker, whose ALB remains
+internal.
+
+Before synthesis, set:
+
+```bash
+export RELEASE_TEST_DRIVER_SECRET_ARN=arn:aws:secretsmanager:us-east-1:613431292675:secret:valkyrie/release-test/package-r-driver-SUFFIX
```
BradenBug, line 332: Should probably be placeholders instead of pointing to a specific account?

docs/RELEASES.md
```diff
@@ -0,0 +1,297 @@
...
+group name, and operator-role ARN. Shared outputs already provide the cluster
+and public subnet IDs. Launches must use those exact values,
+`assignPublicIp=ENABLED`, and only reviewed non-secret command/manifest
+overrides. The default command is the direct-database `release_cli status`
+preflight.
+
+The stage connects to `benchmarks.vals.ai`. Local clients outside the VPC cannot
+call the internal Tracker directly; use the driver for HTTP and database proof.
+
+For migration `e9f0a1b2c3d4`, first prove empty and populated paths in disposable
```
BradenBug: This is ephemeral documentation. Shouldn't live in the repo.

docs/executor-releases/README.md
```diff
@@ -0,0 +1,297 @@
...
+automatic orphan repair in this release-safety layer.
+
+Successive promotions are independent: A, B, and C may all drain concurrently,
+and each retires when its own active execution count reaches zero. There is no
+two-release limit.
+
+Do not delete a release artifact manually while `tracker.release_cli status`
+reports an active-execution or retention blocker.
+
+## Release-test
```
BradenBug, line 308: What is this? There's no context

services/tracker/src/tracker/release_cli.py
```diff
@@ -0,0 +1,73 @@
+"""Operator commands for managing executor releases."""
```
BradenBug: Wait are executor releases managed operationally every time? There's not CI for it?

docs/executor-releases/README.md
OrestesK: An executor release is the immutable executor artifact together with its persisted release identity and lifecycle. PostgreSQL owns the active admission pointer, each benchmark stores initial/current ownership, and each dispatch snapshots the exact artifact URI, digest, and protocol version. This is defined at the top of `docs/RELEASES.md` (lines 1–5), which is the canonical deployment/release contract. I’m keeping that definition centralized here rather than duplicating it across Tracker and Infra docs.

docs/executor-releases/README.md
OrestesK: The concrete account ID is intentional here. This runbook is for the Vals deployment, and repeating the exact target is part of the wrong-account safety boundary; `infra/Makefile` uses the same production account guard. AWS account IDs are identifiers, not secrets. Replacing it with a placeholder would make the operator instructions less verifiable, so I’m keeping the exact ID unless the release-test account policy changes.

services/tracker/src/tracker/database/models.py
```diff
@@ -231,6 +300,11 @@ class Benchmark(SQLModel, table=True):
     finished_at: datetime | None = None
     status: BenchmarkStatus = Field(default=BenchmarkStatus.IN_PROGRESS)
     label: str | None = Field(default=None, index=True)
+    executor_release_id: str | None = Field(default=None, foreign_key="executorrelease.id", index=True)
+    current_execution_release_id: str | None = Field(default=None, foreign_key="executorrelease.id", index=True)
+    executor_artifact_uri: str | None = None
```
BradenBug, line 313: **Structural:** `executor_artifact_uri`/`executor_artifact_digest`/`executor_protocol_version` on `Benchmark` duplicate columns of the immutable `ExecutorRelease` row the benchmark already references via `executor_release_id`. Release rows never mutate after registration, so the FK alone is a stable snapshot — the copies then need the 4-way `benchmark_executor_ownership_complete` CHECK plus hand-propagation at every producer site, and nothing operational reads them: ExecutorHost verifies against the dispatch ledger's own snapshot, and continuation resolves the release row directly. Please drop the three columns and the CHECK; join the (tiny, indexed) release table where a response needs digest/protocol.

Related: the same four-field block is hand-declared in 6 tracker response models + 6 SDK twins and copied field-by-field at 8 constructor sites — one of them (`create_benchmark_table_row`) has no production callers and has already drifted from its live twin in this PR (the live builder gained `error_message`, the dead one didn't). These fields are new here, so the wire shape is still clean-slate: one `ExecutorReleaseProvenance` sub-model per package plus a single `Benchmark` helper collapses ~19 edit sites into 2 and encodes the all-or-nothing invariant in the type instead of the CHECK.

services/tracker/main.py
```diff
@@ -168,6 +178,42 @@ def _taskiq_labels() -> dict[str, str]:
...
+    *,
+    start_benchmark_request_json: dict[str, Any],
+    benchmark_id: UUID,
+    verified_task_ids: list[str],
+) -> None:
+    try:
+        await (
+            process_benchmark.kicker()
+            .with_labels(**_taskiq_labels())
+            .kiq(  # pyright: ignore[reportCallIssue] -- stable host consumes executor_dispatch_id
```
BradenBug: **Structural:** the tracker↔host wire contract has no single definition. The task name, the five `executor_*` kwarg names, `SUPPORTED_PROTOCOL_VERSION`, the stable queue name, the dispatch status strings in the host's raw SQL, and the 64-hex digest check each exist twice — here and in `services/executor_host/supervisor.py` — with nothing pinning the two sides together. This `pyright: ignore` is the tell: the registered `process_benchmark` signature doesn't accept `executor_dispatch_id` (and carries four params its body never reads), so the type checker is silenced at exactly the seam where drift would be caught. A missed rename yields dispatches the host silently skips — the failure mode this PR exists to prevent. Please define the contract once in a small stdlib-only module (frozen payload shape, protocol version, queue name, status values, digest validator), copied into the host image and imported by the tracker; that deletes this ignore and the dead params along with it.

services/executor_host/supervisor.py
OrestesK: Addressed. Terminal recovery now atomically fails every prior `QUEUED`/`RUNNING` dispatch (`dispatch_control.py:61-96`), and executor writes/finalization revalidate the exact `RUNNING` dispatch under benchmark and dispatch row locks (`execution_authority.py:18-43`). The targeted dispatch, Tracker, and ExecutorHost suites pass (84 tests).

services/executor_host/supervisor.py
OrestesK: Addressed. Artifact, digest, spawn, and executor failures now fail the claimed dispatch and atomically move eligible tasks—and the benchmark when no active sibling remains—to `ERROR` (`supervisor.py:254-307`). Unit and PostgreSQL integration coverage exercise those terminal states.

docs/RELEASES.md
OrestesK: Addressed. `docs/RELEASES.md:230-243` now explains that release-test is a dev-sized isolated stage in the production account, defines its resource boundary, and describes the restricted Package R driver used for proof.

docs/executor-releases/README.md
OrestesK: Addressed. `docs/RELEASES.md:237-243` now defines the Package R driver as a static Fargate task definition—not a service—and documents its network, role, logging, secret, and public-IP boundaries.

docs/RELEASES.md
OrestesK: This is intentionally source-controlled operational documentation, not generated run output. `docs/RELEASES.md:230-302` defines the supported release-test deployment and migration contract, so it remains versioned with the implementation it governs.

docs/executor-releases/README.md
OrestesK: Addressed. `docs/RELEASES.md:230-286` now provides the surrounding context: stage/account boundary, Package R driver definition, prerequisites, deployment inputs, launch contract, and the HTTP/database proof purpose.

services/tracker/main.py
OrestesK: Addressed within the selected reliability boundary. START now persists the benchmark mutation and immutable dispatch in one transaction, then publication uses bounded retries and dispatch-state compensation: an unclaimed failure becomes `FAILED`, while a claimed/delivered or superseded dispatch is preserved (`main.py:178-228`, `dispatch_control.py:38-57,190-239`). We deliberately rejected a durable outbox/replay service as disproportionate complexity. The narrow hard-process-death gap remains fail-closed and recoverable through the existing Stop/Retry flow.

services/executor_host/supervisor.py
OrestesK: The cancellation-safety portion is addressed. Task protection is acquired before ownership; claim completion is shielded; cancellation waits for the claim and terminalizes it if ownership was obtained; prelaunch and child failures use the same terminalization path (`supervisor.py:604-660`). We intentionally did not add leases, heartbeats, a reaper, or automatic host-crash recovery. Hard host loss remains fail-closed until the user uses the existing Stop/Retry recovery flow.

services/tracker/src/tracker/database/models.py
OrestesK: Declining this normalization intentionally. Benchmark fields are the immutable initial-admission provenance snapshot, while each dispatch owns its immutable per-invocation snapshot. The benchmark values are consumed by metadata/reporting projections (`models.py:383-396`, `types.py:275-279`), so the premise that nothing reads them is no longer true. Replacing this with FK joins and a new nested public response model would broaden the schema/API contract without changing the approved release behavior; the all-or-none constraint stays.

services/tracker/src/tracker/database/migrations/versions/c7d8e9f0a1b2_add_executor_release_ownership.py
OrestesK: This is handled as a fail-closed first rollout rather than a backward-compatible traffic cutover. The deployment workflow now builds one immutable executor artifact and automatically publishes, verifies, and activates it in the `release-development` / `release-production` jobs. Those jobs run after infrastructure because the sealed activation task and role are created by that deployment. Until first activation, only new starts return 503; existing runs are not reassigned or interrupted. A staged dual-version producer cutover would require migration/rollback machinery that did not exist before this PR and is outside this deployment-release change.

services/tracker/src/tracker/database/migrations/versions/e9f0a1b2c3d4_add_current_execution_release.py
OrestesK: We discussed this and are keeping the migration forward-only. Tracker/PostgreSQL schema rollback was not supported before this PR, so this change does not add an old/new-compatible bridge or a separate migration orchestration system. `docs/RELEASES.md` now states that, after `e9f0a1b2c3d4`, a pre-Package-R Tracker must not be deployed and failures must be fixed forward; database restoration remains a separately approved disaster-recovery action. The generic ECS circuit-breaker rollback is not a supported schema-transition recovery path. Adding database rollback support here would be a new product and operational boundary outside this deployment-focused PR.

results-d01172e1-f598-4992-9477-c31730df129d.json
OrestesK: Yes. The generated results JSON was removed from the effective change and is no longer present in the working-tree status. Validation output now stays under the ignored `.scratch/` workspace rather than becoming a PR artifact.

tests/unit/executor_host/test_supervisor.py
OrestesK: The packaging half is now gated by `.github/workflows/executor-build.yaml`: it runs on an ARM64 runner, verifies both locked service dependency sets, regenerates and checks the ExecutorHost requirements file, builds the immutable PEX, builds both real ARM64 images, and verifies their architectures. The focused ExecutorHost suite also passes locally (22 tests). We intentionally did not add a second service-owned Make/test abstraction solely for this service; the selected CI boundary is the committed-lock plus real-image build, with the existing repository test layout retained.

services/tracker/tests/integration/local/database/test_release_control.py
OrestesK: Addressed. Production routes and PostgreSQL tests now share the canonical operations in `tracker.dispatch_control` and `tracker.utils.run_control`: START uses `admit_start_dispatch`, RETRY/RESUME use `admit_recovery_dispatch`, and Stop uses `apply_stop_benchmark`. The PostgreSQL suite drives those same callables with deterministic transaction barriers, including first activation, promotion/start ordering, recovery, stop, and the pending-Task autoflush regression found during live validation. The final release-control file passes 8/8.

infra/worker_stack.py
OrestesK: We discussed this and deliberately kept the smaller observability boundary. The final change removes the proposed ExecutorHost EMF metrics, claim-age/failure signals, extra alarms, dashboard panels, and MonitoringStack cross-stack Host wiring. Existing ECS rollout health, target tracking, service logs, and the PostgreSQL dispatch ledger remain the operational evidence. Queue-lag and stale-dispatch-age alarms would require new timeout/lease semantics and telemetry ownership, which are not part of the established run behavior or this deployment-release PR.

docs/RELEASES.md
OrestesK: Addressed. `docs/RELEASES.md` now includes `RELEASE_TEST_SANDBOX_PROVIDER_SECRET_ARN` with the other required synthesis inputs, explains the Package R driver and release-test boundary, and provides concrete deployment/live-proof completion checks. `infra/README.md` now lists `driver` as a valid release-test-only scope alongside `shared`, `tracker`, `worker`, `monitoring`, and `all`.

services/tracker/src/tracker/release_cli.py
OrestesK: Addressed. Dev and production now get a dedicated stage-scoped, versioned, retained ExecutorRelease bucket; release-test intentionally uses its isolated test bucket. Registration rejects artifact URIs outside the configured bucket and `releases/` prefix. The sealed release role is the only publisher, writes are create-only through the required `If-None-Match` condition, and Tracker/ExecutorHost retain read-only access to release objects. `infra/executor_release.py` owns build-output publication and activation.

services/tracker/src/tracker/release_cli.py
OrestesK: Addressed. Executor releases are now automated in CI/CD: the deploy workflow builds the immutable ARM64 PEX once per source SHA, carries it as a workflow artifact, then the dev and production release jobs publish, register, verify, and promote it. Production uses the separately gated `production-release` environment. The CLI remains for status and explicit operational rollback/promotion, not as the normal deployment path.

services/tracker/main.py
OrestesK: Addressed. `services/tracker/src/executor_protocol.py` is now the single stdlib-only wire contract for the task name, typed payload keys, supported protocol version, stable queue name, dispatch status values, digest validation, and release-URI validation. Tracker imports it directly; the ExecutorHost image copies that exact module and `supervisor.py` imports the same definitions. The typed `executor_task_signature` replaces the ignored mismatched callable seam, and the Host consumes the immutable fields through `ArtifactDispatch`.

JarettForzano:
Organize the files a bit more by making subfolders. We have close to 6 new files spread across the service that can be organized together

JarettForzano (commented):
instead of just saying executor can we say something like "benchmark_executor"? I want to make it easier to know what it does without looking through the docs.

.github/workflows/maintenance-classification.yaml
JarettForzano, line 1: I didn't understand what this was till I had codex parse the script that it runs, can we rename to "validate-non-disruptive-deployment"

JarettForzano: Also add a description inside

infra/app.py
```diff
@@ -23,6 +27,18 @@
 # Shared infrastructure (VPC, cluster, service discovery, Route53)
 shared = SharedStack(app, stage.stack_id("SharedStack"), stage=stage, env=env)
 
+tracker_repository: aws_ecr.IRepository | None = None
+executor_host_repository: aws_ecr.IRepository | None = None
```
JarettForzano, line 31: Do we havre two different ECR repositories for these?

tests/fixtures/sdk_api/list.json
```diff
@@ -22,6 +22,10 @@
         "agent_name": "sweagent",
         "label": "nightly",
         "model": "claude-sonnet",
+        "executor_release_id": null,
+        "current_execution_release_id": null,
+        "executor_artifact_digest": null,
+        "executor_protocol_version": null,
```
JarettForzano, line 28: We are sending this information to the SDK? Seems like something we would just log

services/tracker/pyproject.toml
```diff
@@ -68,25 +72,24 @@ quote-style = "double"
...
+exclude = ["src/tracker/database/migrations/versions"]
 
 [tool.pytest.ini_options]
 addopts = ["--import-mode=importlib"]
 asyncio_mode = "auto"
-pythonpath = ["."]
-env = [
-    "BROKER_ENVIRONMENT=testing",
-]
+pythonpath = [".", "../.."]
```
JarettForzano, line 91: Move the tests up `../..` instead of adding more surface area to the python path

services/tracker/pyproject.toml
```diff
@@ -58,6 +59,9 @@ allow-direct-references = true
 [tool.hatch.build.targets.wheel]
 packages = ["src/tracker"]
 
+[tool.hatch.build.targets.wheel.force-include]
+"src/executor_protocol.py" = "executor_protocol.py"
+
```
JarettForzano, line 64: Its a pretty small file, you can probably just move it into "src/tracker" rn

services/tracker/main.py
JarettForzano, line 1: Pull all the executor logic into its own file under `utils/`

services/tracker/Dockerfile
JarettForzano, line 1: Root context should just be for something thats at the root, you should add `services/tracker` to the build context and source this file instead of doing the inverse.

services/tracker/tests/unit/test_sandbox.py
```diff
@@ -583,6 +627,7 @@ async def fake_archive_and_upload_output(
             *,
             benchmark_id: str | None = None,
             task_id: str | None = None,
+            execution_is_current: Any = None,
```
JarettForzano: nit: Any type

.github/workflows/deploy.yaml
JarettForzano, line 1: Consolidate duplicated work in this workflow

services/executor_host/requirements.txt
JarettForzano, line 1: Why do we have this if we can just use the uv lock? Also does not look like we have an automation to generate it

services/tracker/src/tracker/utils/resources.py
```diff
@@ -178,7 +178,7 @@ def update_benchmark_resume_arguments(
         arguments = arguments.model_copy(update={"concurrency": concurrency})
 
     benchmark_row.arguments = arguments
-    session.commit()
+    session.add(benchmark_row)
```
JarettForzano, line 181: I get why this was done (to keep it all in one transaction), I suggest coupling the code that needs to be committed together or otherwise you won't be able to guarantee that this stays true in the future

conjfrnk (approved):

infra/app.py
OrestesK: Yes—one repository for the Tracker image and one for ExecutorHost. They have different Dockerfiles/runtime contents and different ECS consumers. We create these repositories for release-test, where both prebuilt images use the same immutable campaign tag; normal dev/prod still use CDK image assets.

tests/fixtures/sdk_api/list.json
OrestesK: Yes, intentionally. These fields are execution provenance, not just telemetry: release ID, artifact digest, and protocol version let SDK consumers identify the exact executor bytes/protocol used for a benchmark. Logs alone would not provide durable per-run attribution.

services/tracker/pyproject.toml
OrestesK: I considered that, but kept it standalone intentionally. Tracker, the executor PEX, and ExecutorHost all consume this contract, while ExecutorHost should not depend on the Tracker package. `executor_protocol.py` stays stdlib-only and is packaged or copied into each runtime.

services/tracker/main.py
OrestesK: I moved the executor domain modules under `tracker/executor/`. I left `_enqueue_executor_dispatch` in `main.py` because it is the HTTP/Taskiq boundary: it translates enqueue outcomes into endpoint-specific HTTP responses. The database and release behavior now lives under `tracker/executor/`, rather than a generic `utils` module.

services/tracker/Dockerfile
OrestesK: Addressed in `29092b14`. Tracker now uses `services/tracker` as both its CDK and CI build context, and its Dockerfile uses context-local `COPY` paths. The root context is limited to ExecutorHost plus the shared protocol. The final ARM64 image build passed.

services/tracker/tests/unit/test_sandbox.py
OrestesK: Addressed in `615eefc2`: this is now `Callable[[], bool] | None`, matching the production callback contract.

services/executor_host/requirements.txt
OrestesK: `uv.lock` is still the source of truth. `requirements.txt` is the generated, hash-pinned input for the slim ExecutorHost image, which installs with `pip --require-hashes` instead of installing `uv`. The executor-build workflow now runs `uv export --locked` and fails if the committed requirements file drifts.

services/tracker/pyproject.toml
OrestesK: The repository-root path is required by the cross-service integration coverage. `test_executor_host_store.py` runs ExecutorHost’s PostgreSQL dispatch store against Tracker’s real schema and disposable PostgreSQL fixtures. Moving that test to the root would also require moving or sharing Tracker’s database test infrastructure, so I kept the explicit path rather than broadening this PR with a test-infrastructure refactor.

PR #607, opened 2026-07-21
Support optional output artifacts

som1shi:
## Summary

Add optional output artifacts without changing task status or scoring when collection fails.

## Changes

- Support `required: false` in SDK and Tracker contracts.
- Skip missing or failed optional artifacts.
- Reject duplicate artifact destination paths.
- Preserve existing required-artifact behavior.
- Document optional trajectories and Model Patch artifacts.

## Related Issues

None.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor
- [x] Docs / config

## Testing

- [ ] Manually tested

- SDK, Tracker, CLI, lint, and type checks pass
- Live Tracker integration has not been run

## Checklist

- [x] Self-reviewed the diff
- [x] Docs updated if needed


services/tracker/src/tracker/sandbox.py
```diff
@@ -634,51 +639,90 @@ async def upload_output_artifacts(
 ) -> None:
     """Upload declared small output artifacts from the sandbox directly to task S3 keys."""
     total_bytes = 0
+    required_artifacts = [artifact for artifact in artifacts if _output_artifact_is_required(artifact)]
+    optional_artifacts = [artifact for artifact in artifacts if not _output_artifact_is_required(artifact)]
 
-    for artifact in artifacts:
+    for artifact in [*required_artifacts, *optional_artifacts]:
```
tthuwng, line 649: **[P1] Reject duplicate artifact destination paths before this reorder.** Required artifacts are now uploaded before optional artifacts, while both use `artifact.path` for the same S3 destination key. The SDK and tracker validators currently allow duplicate normalized paths, so an optional artifact with the same path uploads last and silently overwrites the required artifact. Please reject duplicates at both contract boundaries and add a required/optional collision regression test.

services/tracker/src/tracker/sandbox.py
```diff
@@ -617,11 +621,12 @@ async def _resolve_output_artifact_sandbox_path(sandbox: Sandbox, artifact: Outp
         if source_result.exit_code == _SUCCESS_EXIT_CODE and source_path:
             return source_path
     else:
-        exists = await _exec(sandbox, f"test -f {shlex.quote(source)}")
+        quoted_source = shlex.quote(source)
+        exists = await _exec(sandbox, f"test -f {quoted_source} && ! test -L {quoted_source}")
```
tthuwng: **[P2] This changes existing required-artifact behavior.** Previously a direct required artifact that was a symlink passed `test -f`; this now rejects it, despite the PR summary saying required behavior is unchanged. If this is intentional security hardening, please identify it as a breaking migration and inventory/migrate existing contracts. Otherwise scope the no-symlink behavior to the new optional path and preserve required-artifact compatibility.

tthuwng (changes requested):
Reviewed exact head `565d9a69555ac012e2426e1bd01d72831b750fcd`. Requesting changes for the duplicate-destination integrity issue; the required-artifact symlink compatibility change also needs an explicit decision. The additive `required: false` wire shape and best-effort failure behavior otherwise match the intended scoring-isolation contract. The targeted SDK/tracker/contract suite passed 118 tests.

tthuwng (commented):
Refreshed against current head `8924541e9fcd421fc0fd50a7b8623c448c9aecbe`. The delta from reviewed head `565d9a69555ac012e2426e1bd01d72831b750fcd` only changes `src/valkyrie/cli/s3_config.py` and adds its unit test; it does not touch output-artifact validation or upload ordering. The two existing inline findings remain current and the requested-changes verdict is unchanged.

services/tracker/src/tracker/sandbox.py
som1shi: Fixed in c53037e4. Both SDK and Tracker contract validators now reject duplicate normalized destination paths, including required/optional collisions, with boundary regressions.

services/tracker/src/tracker/sandbox.py
som1shi: Fixed in c53037e4. Required direct artifacts retain the legacy test -f symlink behavior; only optional direct artifacts exclude symlinks. Tests cover required collection and optional skipping.

tthuwng (approved):
Re-reviewed exact head `c77cc852440b3f2bf79c5bdb7a78cb4c953f55e6`. Both prior findings are fixed: duplicate normalized destinations are rejected at both SDK and Tracker boundaries, and required-artifact symlink behavior is preserved. The focused suite passes and ordinary CI is green. The broad optional-artifact catch is consistent with the documented fail-open contract and logs the stack; the optional-symlink documentation note is minor and nonblocking. Approval is code-level; Tracker deployment and the manual live integration remain rollout gates before using `required: false`.

PR #612, opened 2026-07-23
feat: add sandbox scheduler primitives

som1shi:
## Summary

Adds the provider-neutral PostgreSQL scheduler primitives used by [Valkyrie #558](https://github.com/vals-ai/Valkyrie/pull/558). Existing `Task` rows remain the durable queue; this adds no Redis queue, queue table, migration, heartbeat, ownership column, or standalone scheduler service.

## Changes

- Store priority `0..4` and a hashed provider-capacity-pool ID in existing benchmark arguments.
- Select the next `PENDING` task by pool, priority, FIFO order, and run concurrency.
- Serialize capacity checks and task claims with PostgreSQL session advisory locks.
- Transition tasks through `PENDING` → `BUILDING` → `IN_PROGRESS` and reset abandoned builds.
- Keep scheduler-only fields out of the public SDK schema.
- Pin CBS admission support to reviewed [create-benchmark-service #128](https://github.com/vals-ai/create-benchmark-service/pull/128) head `0762c2b26d6366a3ff0c4e2fbecf6a413a145fc3` in both manifests and all three locks.

## Stack

1. [create-benchmark-service #128](https://github.com/vals-ai/create-benchmark-service/pull/128) → `main`
2. **[Valkyrie #612](https://github.com/vals-ai/Valkyrie/pull/612)** → `dev` (this PR)
3. [Valkyrie #558](https://github.com/vals-ai/Valkyrie/pull/558) → currently `ss/scheduler-primitives`; retarget to `dev` after #612 merges
4. [dashboards #4](https://github.com/vals-ai/dashboards/pull/4) → `main`

## Testing

- GitHub `tracker-unit-tests` and `tracker-live-integration-tests` passed on head `499655300c268df7bd8182d3357d5484573de766`.
- PostgreSQL scheduler integration coverage includes priority/FIFO ordering, pool isolation, concurrency, lock reuse/cancellation, abandoned-build recovery, stopped-task admission, and sandbox cleanup.
- CLI tests, SDK compatibility on Python 3.13/3.14, executor build, package build, infrastructure checks, Ruff, BasedPyright, all three lock validations, Codecov patch coverage, and both CDK synth jobs passed on the same head.


services/tracker/src/tracker/scheduler/admission.py
```diff
@@ -0,0 +1,96 @@
...
+                        )
+                    except CapacityObservationUnavailableError:
+                        can_create = True
+                    else:
+                        if not capacity.fits_total(demand):
+                            raise ImpossibleResourceDemandError("Sandbox demand exceeds total Daytona capacity")
+                        can_create = capacity.fits(demand)
+
+                    if can_create:
+                        sandbox = await stack.enter_async_context(create())
```
tthuwng, line 173: [P2] Mark the task BUILDING before sandbox creation. `mark_building()` currently runs only after `create()` has returned, so the task remains PENDING throughout the slow provisioning phase and the dashboard cannot show the real lifecycle state. Move the ownership check/transition to immediately after the capacity grant and before `enter_async_context(create())`; this also avoids allocating and then deleting a sandbox when ownership has already been lost.

tthuwng (commented):
Comment-only review on exact head `4fec8ac`. The scheduler primitives are otherwise cleanly separated; this lifecycle ordering issue is owned by this PR and propagates into the runtime integration in #558.

JarettForzano (changes requested):
We are enforcing all features that go to valkyrie use the CBS client sandboxes and are not hardcoded to any specific provider. Can you make this more generic?

services/tracker/src/tracker/scheduler/gate.py
```diff
@@ -0,0 +1,161 @@
...
+    async def touch(self, ticket: QueueTicket) -> None:
+        """Atomically renew a ticket, restoring lost membership at its original score."""
+        queue_key = self._queue_key(self._prefix(ticket.pool_key), ticket.priority)
+        await self._redis.eval(
+            _RENEW_TICKET,
+            2,
+            queue_key,
+            self._heartbeat_key(ticket),
+            ticket.task_key,
+            repr(ticket.enqueued_at.timestamp()),
```
OrestesK: **[P1] Preserve FIFO when timestamps tie**

The ZSET score is only `enqueued_at.timestamp()`, while `is_next()` selects `ZRANGE 0 0`. Equal-score ZSET members are ordered by member value, so two same-priority tickets with the same timestamp are selected by `task_key` lexical order rather than enqueue order. That breaks the advertised FIFO guarantee.

Verification: on `caa1eaef`, the FIFO test uses distinct timestamps only. A local Valkey probe inserted `z-task` before `a-task` at the same score and returned `a-task` first.

Please add an atomic enqueue-order tie-breaker that survives renewal/recovery, plus an equal-timestamp regression test.

OrestesK (commented):
Verified against exact head `caa1eaef`. One correctness finding, reproduced against local Valkey.

services/tracker/pyproject.toml
```diff
@@ -50,7 +50,7 @@ build-backend = "hatchling.build"
 prerelease = "allow"
 
 [tool.uv.sources]
-create-benchmark-service = { git = "https://github.com/vals-ai/create-benchmark-service.git", rev = "v0.17.3" }
+create-benchmark-service = { git = "https://github.com/vals-ai/create-benchmark-service.git", rev = "c9dfa6593fc88f8465af579efb36de0f42b69c07" }
```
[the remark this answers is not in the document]
som1shi: This is an intentional stack dependency for now: CBS #110 is still open and there is no `v0.18.1` tag. #612 remains blocked on that PR; once it merges and is tagged, this pin should move to the release tag. The PR description now calls this out.

services/tracker/src/tracker/database/models.py
```diff
@@ -214,7 +223,13 @@ def process_bind_param(self, value: BenchmarkArguments | None, dialect: Dialect)
         """Runs when we save the value to the database."""
         if value is None:
             return None
-        return value.model_dump()
+        serialized = value.model_dump(exclude={"priority", "queue_pool_id"})
```
OrestesK, line 242: seems like al lthis does it not serialize None vlaues, prob better pydantic way of doing so

som1shi: `priority` and `queue_pool_id` are intentionally excluded from public `model_dump()` output. `exclude_none=True` cannot re-include excluded fields, so this adapter explicitly persists only their non-null values in the database JSON column.

services/tracker/pyproject.toml
```diff
@@ -51,7 +51,7 @@ build-backend = "hatchling.build"
 prerelease = "allow"
 
 [tool.uv.sources]
-create-benchmark-service = { git = "https://github.com/vals-ai/create-benchmark-service.git", rev = "v0.23.0" }
+create-benchmark-service = { git = "https://github.com/vals-ai/create-benchmark-service.git", rev = "f473b297aac4b8746860fec5674d8ecce8cefc95" }
```
OrestesK: why is this being pinned

services/tracker/src/tracker/database/models.py
```diff
@@ -185,6 +192,8 @@ class BenchmarkArguments(BaseModel):
 
     contract: AgentContractRequest
     concurrency: int
+    priority: int | None = PydanticField(default=None, exclude=True, strict=True, ge=0, le=4)
+    queue_pool_id: str | None = Field(default=None, exclude=True)
```
OrestesK, line 198: why are these excluded

services/tracker/src/tracker/scheduler/admission.py
```diff
@@ -0,0 +1,187 @@
...
+    context: SandboxQueueContext,
+    task_row_id: UUID,
+    expected_started_at: datetime,
+    source: SandboxSource,
+    resources: Resources,
+    create: SandboxFactory,
+) -> Sandbox | None:
+    """Wait for this exact attempt's global turn and enter its sandbox context."""
+    while True:
+        lock = PostgresPoolLock(context.engine, context.pool_id)
```
OrestesK, line 137: waht is locked here, the entire postgres?

services/tracker/pyproject.toml
OrestesK: The temporary pin is explained by CBS #110 still being open, but the stable-deploy merge introduced a second dependency owner that also needs updating.

Tracker is pinned to `f473b297`, while `services/executor_artifact/pyproject.toml` and its lock still select `v0.23.0`. The Executor PEX is built from that artifact lock, and `v0.23.0` does not contain the admission APIs used here (`admission_pool_id` and `check_admission`). The current `executor-build` check fails at artifact lock validation before it can build the PEX.

Can we align the Executor artifact source and lock with Tracker here, then move both to the CBS release tag once #110 is merged?

OrestesK (commented):
Can we also add `services/tracker/src/tracker/scheduler/` to `_EXECUTOR_RELEASE_DIRECTORIES` in `infra/classify_repository_change.py`?

This package is now imported by `process_benchmark()` and shipped inside the immutable Executor PEX, but the release classifier does not recognize it as Executor runtime code. This PR happens to trigger a release through the Tracker dependency change; a later scheduler-only fix could deploy Tracker without building a matching Executor release.

Please add the directory and a classifier test showing that a scheduler source change requires an Executor release.

services/tracker/pyproject.toml
som1shi: This is temporary because CBS #110 is still unreleased. Tracker and the Executor artifact now both use the same current #110 head (`ff4da4c`), and the generated locks agree. After #110 merges and auto-tags, both pins will move to that release tag.

services/tracker/src/tracker/database/models.py
som1shi: These are scheduler storage fields at this layer. `exclude=True` keeps `priority` and `queue_pool_id` out of the public `model_dump()` / SDK wire shape, while `BenchmarkArgumentsType.process_bind_param()` explicitly persists their non-null values in the database JSON. #558 exposes `priority`; `queue_pool_id` stays internal.

services/tracker/src/tracker/scheduler/admission.py
som1shi: Only one provider pool is locked, not PostgreSQL as a whole. `PostgresPoolLock` takes a session advisory lock keyed from that pool ID, so contenders for the same pool serialize while different pool keys continue concurrently. `test_pool_locks_are_isolated_and_reusable` covers that boundary.

services/tracker/pyproject.toml
som1shi: Fixed on the current head. Tracker and `services/executor_artifact` now both resolve CBS `ff4da4c`, all corresponding locks were regenerated, and the current `executor-build` plus root/Tracker lock checks pass. Both temporary SHAs will move to the release tag after CBS #110 merges.

services/tracker/tests/integration/local/database/test_scheduler.py
```diff
@@ -0,0 +1,412 @@
+"""PostgreSQL scheduler behavior tests."""
```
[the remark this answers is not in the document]
som1shi: Fixed in `131eec0a`. The module docstring now includes the exact `uv run pytest tests/integration/local/database/test_scheduler.py` command and a short PostgreSQL scope description. The focused scheduler suite passes (6 tests), and Ruff/format are clean.

services/tracker/tests/integration/local/database/test_scheduler.py
```diff
@@ -0,0 +1,412 @@
...
+)
+import tracker.scheduler.admission as admission
+import tracker.scheduler.store as store
+
+_ATTEMPT = datetime(2026, 7, 27, 12)
+_SOURCE = ImageSource(image="scheduler-test-image")
+_RESOURCES = Resources(vcpu=1, memory=2, disk=3)
+
+
+class _Provider:
```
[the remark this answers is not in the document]
som1shi: Fixed in `131eec0a`. The stand-in is now the public `MockProvider`, and `_context()` uses the renamed class. The focused scheduler suite passes (6 tests), and Ruff/format are clean.

services/executor_artifact/pyproject.toml
```diff
@@ -9,5 +9,5 @@ dependencies = ["tracker"]
 prerelease = "allow"
 
 [tool.uv.sources]
-create-benchmark-service = { git = "https://github.com/vals-ai/create-benchmark-service.git", rev = "v0.27.6" }
```
OrestesK: remember to unpin before merging

services/tracker/src/tracker/database/models.py
OrestesK, line 198: why dont we want these values in the public model dump

services/executor_artifact/pyproject.toml
som1shi: Ack

services/tracker/src/tracker/database/models.py
som1shi: theyre not user facing so like extraneous
priority is already public otherwise, queue pool id is like irrelevant

OrestesK (approved):

JarettForzano (approved):

PR #629, opened 2026-07-29
Pass agent environment to sandbox commands

tthuwng:
## What changed

- Pass Tracker's already-resolved task environment to agent dependency installation and agent execution through the sandbox command API.
- Keep values out of shell command text.
- Add regression coverage for both command stages and the task-execution handoff.

## Why

Masscan is routed to a targeted Daytona Linux VM. Unlike container sandboxes, that VM does not expose `SandboxCreateRequest.env_vars` to newly created PTYs, so `setup.sh` failed with `MODEL_PROXY_SSH_KEY is required`.

CBS v0.22.1 attempted to restore sandbox-wide inheritance with `update_env`, but the required live smoke against the exact Masscan snapshot failed with `DaytonaNotFoundError: Failed to update environment: 404 page not found`. This PR uses the PTY command environment path that the runner actually supports and limits secret lifetime to dependency and agent commands.

## Validation

- Exact `us-west-3` Masscan snapshot live smoke: dependency setup and agent execution both received a surrogate variable; direct commands did not retain it; cross-target lookup passed; sandbox deleted.
- Local Docker Tracker API and worker built from this branch; `/health` returned `{"status":"ok"}`.
- `make test`: 420 passed, Alembic passed, 86.94% coverage.
- Ruff lint and format passed.
- Basedpyright reports only two pre-existing errors at unchanged `task_execution.py` lines 516 and 720.

### DIFF
--- a/docs/CONTRACTS.md
+++ b/docs/CONTRACTS.md
@@ -208,7 +208,7 @@ Omit this field, or set it to an empty list, to keep unrestricted sandbox egress
 
 ### `secrets: dict`
 
-Secrets required by the agent. Maps environment variable names to AWS Secrets Manager secret names. Tracker resolves them and injects them only during agent dependency installation and execution. Raw values are never stored.
+Secrets specified in the contract or CLI are used when installing dependencies and running the agent. Contract entries map environment variable names to AWS Secrets Manager secret names. Raw values are never stored.
 
 ```yaml
 secrets:
--- a/docs/CONTRACTS.md
+++ b/docs/CONTRACTS.md
@@ -208,7 +208,7 @@ Omit this field, or set it to an empty list, to keep unrestricted sandbox egress
 
 ### `secrets: dict`
 
-Secrets required by the agent. Maps environment variable names to AWS Secrets Manager secret names. These are resolved at sandbox creation time - raw values are never stored.
+Secrets required by the agent. Maps environment variable names to AWS Secrets Manager secret names. Tracker resolves them and injects them only during agent dependency installation and execution. Raw values are never stored.
 
 ```yaml
 secrets:
--- a/services/tracker/src/tracker/utils/task_execution.py
+++ b/services/tracker/src/tracker/utils/task_execution.py
@@ -569,11 +569,15 @@ async def _process_task_attempt(
         if benchmark_started_by_email:
             identity["email"] = benchmark_started_by_email
 
-        env_vars = {
-            **resolve_secrets(start_benchmark_request.contract.secrets, harness_config.aws),
+        agent_secrets = resolve_secrets(start_benchmark_request.contract.secrets, harness_config.aws)
+        agent_secrets.pop("DAYTONA_SANDBOX_OTEL_EXTRA_LABELS", None)
+        agent_env_vars = {
+            **agent_secrets,
             "RUN_ID": str(benchmark_id),
             "TASK_ID": task_row.task_id,
             "IDENTITY": json.dumps(identity),
+        }
+        sandbox_env_vars = {
             # Tags sandbox-internal OTel telemetry with our IDs + environment so traces/logs/metrics
             # are filterable per benchmark run and separable from other environments sharing the
             # same Daytona account (sandbox OTLP export is account-level).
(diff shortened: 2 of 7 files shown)


JarettForzano (approved):

JarettForzano (dismissed):

docs/CONTRACTS.md
```diff
@@ -208,7 +208,7 @@ Omit this field, or set it to an empty list, to keep unrestricted sandbox egress
 
 ### `secrets: dict`
 
-Secrets required by the agent. Maps environment variable names to AWS Secrets Manager secret names. These are resolved at sandbox creation time - raw values are never stored.
+Secrets required by the agent. Maps environment variable names to AWS Secrets Manager secret names. Tracker resolves them and injects them only during agent dependency installation and execution. Raw values are never stored.
```
JarettForzano: "Secrets specified in the contract or the CLI are used when installing dependencies and running the agent"

services/tracker/src/tracker/utils/task_execution.py
```diff
@@ -569,17 +569,13 @@ def task_is_stopped() -> bool:
         if benchmark_started_by_email:
             identity["email"] = benchmark_started_by_email
 
-        env_vars = {
-            **resolve_secrets(start_benchmark_request.contract.secrets, harness_config.aws),
+        agent_secrets = resolve_secrets(start_benchmark_request.contract.secrets, harness_config.aws)
+        agent_secrets.pop("DAYTONA_SANDBOX_OTEL_EXTRA_LABELS", None)
```
JarettForzano: why do we pop the secret here?

services/tracker/src/tracker/utils/task_execution.py
```diff
@@ -591,7 +587,12 @@ def task_is_stopped() -> bool:
             sandbox_name=task_row.task_id,
             source=task_data.source,
             labels=labels,
-            env_vars=env_vars,
+            # Sandbox telemetry is configuration; agent secrets stay command-scoped.
```
JarettForzano: keep the original comment
```
# Tags sandbox-internal OTel telemetry with our IDs + environment so traces/logs/metrics
            # are filterable per benchmark run and separable from other environments sharing the
            # same Daytona account (sandbox OTLP export is account-level).
````

File: services/tracker/src/tracker/utils/task_execution.py [hunk 4]
```diff
@@ -569,17 +569,13 @@ def task_is_stopped() -> bool:
         if benchmark_started_by_email:
             identity["email"] = benchmark_started_by_email
 
-        env_vars = {
-            **resolve_secrets(start_benchmark_request.contract.secrets, harness_config.aws),
+        agent_secrets = resolve_secrets(start_benchmark_request.contract.secrets, harness_config.aws)
```
JarettForzano: nit: keep same style `` **resolve_secrets(start_benchmark_request.contract.secrets, harness_config.aws),``

JarettForzano (commented):
Ensure agent secrets are only injected into the command when we are calling the agent or installing dependencies

docs/CONTRACTS.md
tthuwng: Updated the contract docs with this wording and retained the contract mapping and raw-value storage details.

services/tracker/src/tracker/utils/task_execution.py
tthuwng: Removed the special-case pop. Resolved contract and CLI secrets are now assembled directly into the agent-only environment used for dependency installation and agent execution.

services/tracker/src/tracker/utils/task_execution.py
tthuwng: Restored the original sandbox telemetry comment at the point where the Daytona-only OTel label is created.

services/tracker/src/tracker/utils/task_execution.py
tthuwng: Inlined resolve_secrets into agent_env_vars to keep the original style and remove the one-use mutable value.

PR #630, opened 2026-07-30
Measure agent command duration in tracker process

nikil-ravi:
## Summary
`stream_command_output` timed agent commands by wrapping them in the sandbox to write `date +%s%N` into `.start_ns`/`.end_ns` files under `/tmp/.valkyrie`, then `cat`-ing both files and doing `int(end_ns) - int(start_ns)`. It never checked that the `cat` succeeded, so when `end_ns` was absent, the `cat` error text (`cat: ...end_ns: No such file or directory`) was passed straight to `int()`, producing a misleading `ValueError` that surfaced as the task error.

This hit 4/200 tasks (all `revprotocol`). The failing task's live CloudWatch stream confirms it, ending with:
`[ERROR] invalid literal for int() with base 10: 'cat: /tmp/.valkyrie/8d0783cb97c542e1b2e4448e7d98a8fd.end_ns: No such file or directory'`
Because duration was computed before artifact upload, those tasks have no console/trajectory artifacts in S3. The temp files most likely disappeared due to the agent removing/replacing `/tmp/.valkyrie` or a command-session interruption near completion.

Fix: keep the in-sandbox `start_ns`/`end_ns` timing (it deliberately excludes the tracker→sandbox request round-trip and program cold-start, ~5-6s, so the measurement stays accurate), but make the read non-fatal — check the `cat` exit codes and only `int()` on success, degrading to a tracker-side `time.monotonic()` fallback when the files are missing/unparseable instead of crashing.

## Changes
- `services/tracker/src/tracker/sandbox.py`: extract the timing read into `_read_sandbox_duration`, which returns the sandbox-measured duration when both `cat`s succeed and parse, and otherwise returns the `time.monotonic()` fallback captured around the command stream. Timing files are still written/cleaned up as before.
- `services/tracker/tests/unit/test_sandbox.py`: assert the sandbox-embedded timing path (duration from files, temp files removed) and add a fallback test where a missing `end_ns` no longer raises `ValueError`.

## Related Issues
Closes #

## Type of Change
- [x] Bug fix

## Testing
- [x] Manually tested

`uv run pytest tests/unit/test_sandbox.py` → 45 passed; ruff + basedpyright clean.

## Checklist
- [x] Self-reviewed the diff
- [ ] Docs updated if needed

Link to Devin session: https://app.devin.ai/sessions/6d4830d12fa44c64981b3080f5349545
Requested by: @nikil-ravi

---

<a href="https://app.devin.ai/review/vals-ai/valkyrie/pull/630" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/services/tracker/src/tracker/sandbox.py
+++ b/services/tracker/src/tracker/sandbox.py
@@ -405,6 +405,7 @@ async def install_agent_dependencies(
 _TIMEOUT_EXIT_CODE: int = 124
 _OS_KILL_EXIT_CODE: int = 137
 _SUCCESS_EXIT_CODE: int = 0
+_STATUS_DIR = "/tmp/.valkyrie"
 _OUTPUT_TAIL_MAX_CHARS = 64 * 1024
 _EGRESS_RETRY = retry(
     retry=retry_if_exception_type(ProviderSandboxError) & retry_if_not_exception_type(SandboxNotFoundError),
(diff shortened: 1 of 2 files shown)


JarettForzano (approved):

JarettForzano:
Add fallback

PR #642, opened 2026-08-03
Forward benchmark sandbox secret references to sandbox providers

Chen-Oliver:
## Summary
- Plumbs `RetrieveTaskResponse.sandbox_secrets` (env var name -> Daytona org-secret name, reference-only) from task retrieval through `create_sandbox` into the provider `SandboxCreateRequest`
- Rejects env var names declared as both plaintext `env_vars` and provider-managed `sandbox_secrets`, raising `InvalidSandboxConfigurationError` — a deterministic config error that fails the attempt immediately rather than triggering the fresh-sandbox `SandboxSetupError` retry, which cannot help
- Companion to vals-ai/create-benchmark-service#127

## Depends on
Requires vals-ai/create-benchmark-service#127. The `create-benchmark-service` pin in `services/tracker` and `services/executor_artifact` (plus the lockfiles) currently points at that PR's commit so CI runs against the dependency this branch needs; swap it to the release tag once #127 merges and a release is cut, mirroring how #633 paired the volumes plumbing with its dependency bump.

## Test plan
- `test_create_sandbox_passes_request_to_provider` asserts secrets reach the provider request
- `test_create_sandbox_rejects_plaintext_and_secret_environment_collision` asserts the fail-fast path (passes today; raises before the request is built)
- `test_process_task_forwards_native_secret_references_without_resolving_values` asserts references are forwarded untouched while legacy contract secrets still resolve to plaintext, with no cross-contamination

---

<a href="https://app.devin.ai/review/vals-ai/valkyrie/pull/642" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

## Test results
- `test_create_sandbox_rejects_plaintext_and_secret_environment_collision` passes locally (fail-fast path, raises before the provider request is built)
- The two forwarding tests fail as expected until the create-benchmark-service pin is bumped past vals-ai/create-benchmark-service#127
- Live Daytona validation of the provider-side reference path (tenant keys can attach native secret references at sandbox creation) is documented on vals-ai/create-benchmark-service#127

## Live end-to-end validation
With the tracker, executor artifact, and benchmark service all built against the vals-ai/create-benchmark-service#127 commit (pins temporarily overridden locally), a full `valk run start` against real Daytona created the generation sandbox with the task-declared secret reference, and the benchmark service verified in-sandbox injection of the env var as an opaque placeholder (observation: `PRESENT 27 OPAQUE`). The task finished and the run scored 100. With the pin now committed at that revision, the full tracker unit suite (465 tests) and the live integration suite pass in CI.

### DIFF
--- a/services/executor_artifact/pyproject.toml
+++ b/services/executor_artifact/pyproject.toml
@@ -9,5 +9,5 @@ dependencies = ["tracker"]
 prerelease = "allow"
 
 [tool.uv.sources]
-create-benchmark-service = { git = "https://github.com/vals-ai/create-benchmark-service.git", rev = "v0.23.0" }
+create-benchmark-service = { git = "https://github.com/vals-ai/create-benchmark-service.git", rev = "cfda9455d9e35533c023ed7f20980bb4c8358a78" }
 tracker = { path = "../tracker" }
--- a/services/executor_artifact/pyproject.toml
+++ b/services/executor_artifact/pyproject.toml
@@ -9,5 +9,5 @@ dependencies = ["tracker"]
 prerelease = "allow"
 
 [tool.uv.sources]
-create-benchmark-service = { git = "https://github.com/vals-ai/create-benchmark-service.git", rev = "cfda9455d9e35533c023ed7f20980bb4c8358a78" }
+create-benchmark-service = { git = "https://github.com/vals-ai/create-benchmark-service.git", rev = "v0.26.0" }
 tracker = { path = "../tracker" }
--- a/services/tracker/src/tracker/sandbox.py
+++ b/services/tracker/src/tracker/sandbox.py
@@ -161,6 +161,18 @@ def _set_sandbox_span_attributes(sandbox: Sandbox) -> None:
     span.set_attribute("valkyrie.sandbox_state", sandbox.state)
 
 
+def _reject_plaintext_secret_collisions(
+    env_vars: dict[str, str] | None,
+    sandbox_secrets: dict[str, str] | None,
+) -> None:
+    overlapping_env_names = sorted(set(env_vars or {}) & set(sandbox_secrets or {}))
+    if overlapping_env_names:
+        raise InvalidSandboxConfigurationError(
+            "Sandbox environment variables cannot be both plaintext and provider-managed secrets: "
+            f"{', '.join(overlapping_env_names)}"
+        )
+
+
 @logfire.instrument("sandbox.create", extract_args=False)
 async def _create_sandbox(
     provider: SandboxProvider,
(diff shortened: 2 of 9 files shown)


services/tracker/src/tracker/utils/task_execution.py
```diff
@@ -658,6 +658,7 @@ def task_is_stopped() -> bool:
             source=task_data.source,
             labels=labels,
             env_vars=env_vars,
+            sandbox_secrets=task_data.sandbox_secrets,
```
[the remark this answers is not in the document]
Chen-Oliver: Known and deliberate for now: the create-benchmark-service pin bump is deferred until vals-ai/create-benchmark-service#127 is released. The PR description documents the two tests that stay red until then. The pin and lockfiles will be bumped on this branch before merge.

Chen-Oliver:
Final disposition of the Greptile findings: the collision check raises InvalidSandboxConfigurationError (not the retried SandboxSetupError), so a deterministic misconfiguration fails the attempt immediately instead of burning the fresh-sandbox retry. Beyond that, error handling and metrics are intentionally unchanged (049df1d5): the error flows through the existing generic handler and the sandbox.create.errors metric, same as any other SandboxError.

services/executor_artifact/pyproject.toml
```diff
@@ -9,5 +9,5 @@ dependencies = ["tracker"]
 prerelease = "allow"
 
 [tool.uv.sources]
-create-benchmark-service = { git = "https://github.com/vals-ai/create-benchmark-service.git", rev = "v0.23.0" }
+create-benchmark-service = { git = "https://github.com/vals-ai/create-benchmark-service.git", rev = "cfda9455d9e35533c023ed7f20980bb4c8358a78" }
```
BradenBug: reminder to repin

services/tracker/src/tracker/sandbox.py
```diff
@@ -169,8 +170,15 @@ async def _create_sandbox(
     labels: dict[str, str] | None = None,
     env_vars: dict[str, str] | None = None,
     volumes: list[VolumeMount] | None = None,
+    sandbox_secrets: dict[str, str] | None = None,
 ) -> Sandbox:
     """Create a sandbox through its provider."""
+    overlapping_env_names = sorted(set(env_vars or {}) & set(sandbox_secrets or {}))
+    if overlapping_env_names:
```
BradenBug: nit: would put this and 176 in helper method

BradenBug (approved):

services/executor_artifact/pyproject.toml
Chen-Oliver: Repinned to v0.26.0 in 02341e89 (both pyprojects and all three lockfiles).

services/tracker/src/tracker/sandbox.py
Chen-Oliver: Extracted into _reject_plaintext_secret_collisions in 02341e89.

PR #655, opened 2026-08-07
fix: close benchmark service client

JarettForzano:
## Summary
Close the benchmark-service client after `/start-benchmark` preflight validation to release its HTTP connection pool and cached sandbox providers.

## Changes
- Wrap benchmark-service health and task verification in cleanup handling.
- Preserve the existing error response when client cleanup itself fails.
- Extend the existing unreachable-service test to verify cleanup is awaited on the failing path.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor
- [ ] Docs / config

## Testing
- [ ] Manually tested

## Checklist
- [x] Self-reviewed the diff
- [x] Docs updated if needed; no documentation changes are required

### DIFF
--- a/services/tracker/main.py
+++ b/services/tracker/main.py
@@ -408,23 +408,29 @@ async def start_benchmark(
     # Validate benchmark service is reachable + tasks resolve BEFORE creating the DB row,
     # so failed auth / unreachable services don't pollute the benchmark list.
     try:
-        _ = await benchmark_service.health_check()
-    except httpx.ConnectError as exc:
-        raise HTTPException(
-            status_code=502,
-            detail=f"Benchmark service '{request.benchmark_name}' is not reachable",
-        ) from exc
+        try:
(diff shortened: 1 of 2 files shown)


OrestesK:
question: Could we make client ownership harder to get wrong instead of fixing only this call site?

This change closes the client in the right place and fixes this leak. But each caller still has to remember to close its client. Since `BenchmarkServiceClient` already supports `async with`, would it make sense to use that pattern consistently for short-lived clients?

We would need to preserve the behavior here where a cleanup failure does not hide the original result or error, because the client’s current `__aexit__()` does not handle that for us.

If that is too broad for this PR, a follow-up is reasonable. For this PR, could we keep explicit tests for cleanup after a successful preflight and for a cleanup failure not changing the response? The current assertion only covers the unreachable-service path.

tthuwng (approved):

PR #656, opened 2026-08-08
refactor(storage): abstract object operations

OrestesK:
## Purpose

Make object storage replaceable without changing Tracker's current S3-backed behavior. This is the root PR of the stack: later PRs depend on its provider-neutral storage boundary.

## What changes

- Introduces AWS-free object-storage and artifact-location contracts in `tracker.runtime`.
- Moves artifact key construction, agent-bundle copy-if-absent behavior, and agent listing/filtering to provider-neutral helpers.
- Keeps `AWSRuntime` and its already-selected AWS client provider as the S3 composition authority.
- Adapts the existing S3 implementation to the new contracts; benchmark, task, sandbox, reporting, API, and CLI flows consume the boundary rather than constructing storage clients themselves.

## Behavior preserved

- Existing S3 key formats and artifact locations.
- Presigned URL expiry and download behavior.
- Multipart upload behavior, retry/error translation, and telemetry instrumentation.
- Existing AWS credential/runtime selection; no credential object moves onto domain interfaces.

## Deliberate non-goals

- No local storage provider is added in this PR.
- No generic provider registry, selector, or factory is introduced.
- No API payload, route, or user-visible artifact workflow changes.

## Review guide

1. Start with the `tracker.runtime` storage and artifact contracts/helpers.
2. Verify the S3 adapter remains a thin implementation behind the selected runtime client.
3. Follow the migrated artifact call sites to confirm existing behavior is reached through the new boundary.
4. Review the focused S3 and artifact-policy tests for copy/list/download semantics.

## Stack and merge order

1. **#656 Storage** (`dev` ← `ok/runtime-storage`)
2. #657 Secrets (`ok/runtime-storage` ← `ok/runtime-secrets`)
3. #658 Logging (`ok/runtime-secrets` ← `ok/runtime-logs`)
4. #659 Executor artifacts (`ok/runtime-logs` ← `ok/executor-artifacts`)

Merge in that order after #656's required approval.

## Validation

- 690 Tracker unit tests passed (one existing Starlette/httpx warning).
- 45 executor-host and end-to-end telemetry tests passed.
- Basedpyright, Ruff, formatting, diff checks, targeted LSP, and independent review passed.
- Local executable diff-coverage estimate: **89.51%** (above Codecov's #656 patch target).

## Live validation

- `uv run pytest tests/integration/live/api/test_s3_routes.py` against `ValkDevSharedStack`: **2 passed** (one existing Starlette/httpx warning).
[...]

PR #657, opened 2026-08-08
refactor(secrets): abstract secret resolution

OrestesK:
## Purpose

Make secret resolution replaceable while preserving Tracker's existing explicit-key, managed-IAM, and maintenance-cleanup authority rules. This PR builds on #656's storage boundary.

## What changes

- Introduces AWS-free `SecretStore` and secret-reference resolution contracts.
- Makes `SecretsManagerStore` a thin adapter over the already-selected `AWSClientProvider`.
- Routes request, worker, orchestration, notification, stop, and maintenance-cleanup paths through resolved runtime authority.
- Keeps the default-chain decision visible at the maintenance-cleanup composition root rather than hiding it behind a domain interface.

## Behavior preserved

- Explicit caller-provided credentials remain authoritative where they were before.
- Managed-IAM/runtime credential selection remains authoritative where it was before.
- Existing secret-reference names, Secrets Manager lookup behavior, and failure behavior remain unchanged.
- Notifications, stop/cleanup work, and orchestration retain their existing authorization flow.

## Deliberate non-goals

- No credential values or AWS client types are exposed through domain contracts.
- No generic provider registry, selector, factory, or capability bundle is added.
- No new secret backend is shipped; the seam enables a future local backend without making one a runtime requirement.

## Review guide

1. Read the runtime secret contract and `SecretsManagerStore` adapter.
2. Trace each authority decision from request or worker composition to the secret lookup.
3. Confirm maintenance cleanup is the only default-chain composition point.
4. Review tests for explicit credentials, managed runtime authority, notification, and cleanup paths.

## Stack and merge order

1. #656 Storage (`dev` ← `ok/runtime-storage`)
2. **#657 Secrets** (`ok/runtime-storage` ← `ok/runtime-secrets`)
3. #658 Logging (`ok/runtime-secrets` ← `ok/runtime-logs`)
4. #659 Executor artifacts (`ok/runtime-logs` ← `ok/executor-artifacts`)

Merge only after #656 merges; then continue in stack order.

## Validation

- 690 Tracker unit tests passed (one existing Starlette/httpx warning).
- 45 executor-host and end-to-end telemetry tests passed.
- Basedpyright, Ruff, formatting, diff checks, targeted LSP, and independent review passed.
- The final stack's Codecov-sensitive storage and logging changes have local executable diff-coverage estimates above their required patch targets.

## Live validation scope

- The stack ran `uv run pytest tests/integration/live/api/test_s3_routes.py` against `ValkDevSharedStack`: **2 passed** (one existing Starlette/httpx warning).
- That probe is direct evidence for the storage route inherited from #656; it is not a live Secrets Manager retrieval test.
- Secrets Manager retrieval, sandbox orchestration, CloudWatch writes, and sealed executor-release activation remain covered by the green CI and unit contracts.

### DIFF
--- a/services/tracker/src/tracker/runtime/secrets.py
+++ b/services/tracker/src/tracker/runtime/secrets.py
@@ -1,0 +1,35 @@
+"""Provider-neutral secret reads and reference resolution."""
+
+from typing import Protocol, TypeAlias
+
+from tracker.exceptions import SecretsError
+
+
+JsonScalar: TypeAlias = str | int | float | bool | None
+JsonValue: TypeAlias = JsonScalar | list["JsonValue"] | dict[str, "JsonValue"]
+SecretValue: TypeAlias = JsonValue
+
+
+class SecretStore(Protocol):
+    """Synchronous access to named secret values."""
+
+    def get(self, name: str) -> SecretValue:
+        """Return decoded JSON, or the raw string when the value is not JSON."""
+        raise NotImplementedError
+
+
+def resolve_secrets(secrets: dict[str, str], secret_store: SecretStore) -> dict[str, str]:
+    """Resolve environment-variable secret references to their current values."""
+    if not secrets:
+        return {}
+
+    resolved: dict[str, str] = {}
+    for env_name, secret_name in secrets.items():
+        secret_value = secret_store.get(secret_name)
+        if isinstance(secret_value, dict):
+            if env_name not in secret_value:
+                raise SecretsError(f"Key '{env_name}' not found in JSON secret '{secret_name}'")
+            resolved[env_name] = str(secret_value[env_name])
+        else:
+            resolved[env_name] = str(secret_value)
+    return resolved
--- a/services/tracker/main.py
+++ b/services/tracker/main.py
@@ -45,7 +45,7 @@ from tracker.aws.resolver import (
     resolve_run_aws_runtime_and_access_key_config,
     resolve_start_aws_runtime,
 )
-from tracker.aws.secrets import resolve_secrets
+from tracker.aws.secrets import SecretsManagerStore
 from tracker.agent.contract import get_contract_from_zip_bytes
 from tracker.aws.s3 import (
     S3_BENCHMARKS_PREFIX,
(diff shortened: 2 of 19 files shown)


services/tracker/src/tracker/runtime/secrets.py
```diff
@@ -0,0 +1,35 @@
...
+
+    resolved: dict[str, str] = {}
+    for env_name, secret_name in secrets.items():
+        secret_value = secret_store.get(secret_name)
+        if isinstance(secret_value, dict):
+            if env_name not in secret_value:
+                raise SecretsError(f"Key '{env_name}' not found in JSON secret '{secret_name}'")
+            resolved[env_name] = str(secret_value[env_name])
+        else:
+            resolved[env_name] = cast(str, secret_value)
```
BradenBug: When secret_value is not a dict, `cast(str, secret_value)` doesn't convert the value - it just silences the type checker. If the secret is a JSON number, boolean, null, or array, the returned dict will contain non-string values. The test `test_resolve_secrets_preserves_non_object_values` confirms this behavior. Either convert to string with `str(secret_value)` or change the return type to `dict[str, SecretValue]`.

BradenBug (approved):

OrestesK:
## Review feedback disposition

- Fixed the unresolved secret-value coercion feedback in `3c7a0233`: non-object decoded JSON values now cross the environment/header boundary through `str(...)`, matching `dict[str, str]`.
- The focused secret tests cover raw strings, lists, numbers, booleans, and null; 9 focused tests and 690 Tracker unit tests pass locally.
- Earlier Codecov and superseded bot reviews predate this head.

services/tracker/src/tracker/runtime/secrets.py
OrestesK: Fixed in 3c7a0233: non-object decoded JSON secret values now cross the environment/header boundary through str(...), and focused coverage asserts the resulting strings.

OrestesK:

- #656
    - #657 ◀
        - #658
            - #659

PR #664, opened 2026-08-10
fix(tracker): stop forced sandbox teardown from racing live executor work #patch

conjfrnk:
## What

- Force stop gives a live executor a bounded 15s window to tear down its own sandboxes, then deletes only the stragglers.
- No active dispatch means nothing can be running inside them, so the window is skipped and the path is unchanged.
- The orphan reaper now emits the same `sandbox.delete` record as every other caller, as `initiated_by="orphan_cleanup"`.

## Why

- Force stop deleted every labelled sandbox in the same request that stopped the tasks. The executor tears down those same sandboxes on its next poll, so deletes landed underneath commands still in flight and failed live work.
- Bounded, not open-ended: a stalled executor can't hold its sandboxes past the window, so force stop keeps its kill guarantee. 15s fits inside the 60s ALB idle timeout.
- `sandbox_cleanup.py` was the last caller deleting through the provider directly, so its deletions were the ones no log could attribute. It still calls the provider directly on purpose — the audited wrapper swallows `SandboxNotFoundError` and generic failures, which would break the Lambda's outcome counts.

## How tested

- 3 new unit cases: a released sandbox is never deleted here; with no dispatch it's deleted on the first pass; one never released is still reaped at expiry. The first fails on `dev`.
- Cleanup test pins one `orphan_cleanup` record per delete attempt, across `deleted`, `already_gone` and `failed`.
- `services/tracker && make test`: 575 passed, 88.34% coverage (85% gate). Root `make test`: 204 passed. `ruff` and `basedpyright` clean.
- Live suite is manual-only on non-prod PRs. `test_force_stop_end_to_end` still passes, exercising window expiry rather than executor release, since its blocking `TestClient` call starves the executor loop.

### DIFF
--- a/services/tracker/src/tracker/utils/run_control.py
+++ b/services/tracker/src/tracker/utils/run_control.py
@@ -1,5 +1,7 @@
 """Operations that stop, resume, or retry a run and tear down its sandboxes."""
 
+import asyncio
+import time
 import traceback
 from collections.abc import AsyncGenerator
 from datetime import datetime
@@ -23,7 +25,7 @@ from tracker.database.models import (
     Task,
     TaskStatus,
 )
-from tracker.executor.dispatch_control import terminalize_active_dispatches
+from tracker.executor.dispatch_control import active_dispatch_exists, terminalize_active_dispatches
 from tracker.exceptions import TrackerServiceError
 from tracker.logging import get_logger
 from tracker.sandbox import delete_sandbox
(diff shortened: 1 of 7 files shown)


services/tracker/src/tracker/utils/run_control.py
```diff
@@ -142,12 +168,16 @@ async def force_stop_sandboxes(
...
+    executor_active = active_dispatch_exists(session, benchmark_row.id)
+
     session.commit()
 
     # Iterate through each running sandbox and stop it, collecting error messages
     results: dict[str, str | None] = {}
     try:
-        async for sandbox in sandbox_generator(benchmark_row, provider, task_ids=task_ids):
+        drain_seconds = SANDBOX_DRAIN_TIMEOUT_SECONDS if executor_active else 0.0
+        for sandbox in await drain_sandboxes(benchmark_row, provider, task_ids, drain_seconds):
```
OrestesK, line 180: async gone?

OrestesK (approved):

PR #667, opened 2026-08-11
test(tracker): unflake the output-artifact error log assertion #patch

conjfrnk:
## What

- `test_required_missing_output_artifact_persists_compatible_error` now waits for the log flush that carries the error it asserts on, instead of the first flush of any kind.

## Why

- `buffer_logs` dispatches every write with `loop.run_in_executor(...)` and never awaits it, so the flush carrying `[ERROR] ...` can land after `run_process_task` returns.
- The test set its event on the *first* message to arrive and then asserted a *specific* one was present, so any earlier flush released the wait too soon and the assertion read a list that did not have the error line yet.
- It fails as a bare `assert False`, which reads like a product bug rather than a test race. Seen on `cjf/force-stop-sandbox-drain` twice in a row and on `ok/runtime-storage` on Aug 9; it does not reproduce on a developer machine, only under CI's thread contention.
- Timeout raised 1s → 10s so a genuinely missing log fails as a timeout naming the wait, not as `assert False`.

## How tested

- `services/tracker && make test`: 575 passed, coverage 88.34% (gate 85%).
- The assertion is unchanged in substance — same message, same list — so a real regression in the error log still fails the test.

---

<a href="https://app.devin.ai/review/vals-ai/valkyrie/pull/667" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/services/tracker/tests/unit/utils/test_benchmark_service_failures.py
+++ b/services/tracker/tests/unit/utils/test_benchmark_service_failures.py
@@ -179,8 +179,10 @@ class TestBenchmarkServiceFailures:
         start_benchmark_request, task_row, benchmark_id, authority = create_task_environment(
             contract, database_session, harness_config
         )
+        expected_error = "Output artifact error: Required output artifact missing: /tmp/valkyrie/artifacts/missing.json"
+        expected_log = f"[ERROR] {expected_error}"
         logged_messages: list[str] = []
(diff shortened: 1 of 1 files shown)


conjfrnk:
Superseded: the same fix is in #664 (`3621a067`), which is already approved, so it lands with that merge instead of waiting on a second review.

PR #670, opened 2026-08-12
chore(infra): source imported secret names from deployment configuration

lgnashold:
## Changes
-

## Type of Change
- [x] Docs / config

## Testing

## Checklist
- [x] Self-reviewed the diff
- [x] Docs updated if needed

Link to Devin session: https://app.devin.ai/sessions/43a9c6c0c5f746cdb5d0c978323296cf
Requested by: @lgnashold

---

<a href="https://app.devin.ai/review/vals-ai/valkyrie/pull/670" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -701,6 +701,10 @@ jobs:
           steps.freshness.outputs.current == 'true' &&
           needs.classify-deployment.outputs.executor_stack_deploy_required == 'true'
         env:
+          AUTH_REQUIRED: ${{ secrets.AUTH_REQUIRED }}
+          DESCOPE_MANAGEMENT_KEY_SECRET_NAME: ${{ secrets.DESCOPE_MANAGEMENT_KEY_SECRET_NAME }}
+          DESCOPE_PROJECT_ID: ${{ secrets.DESCOPE_PROJECT_ID }}
+          SENTRY_DSN_SECRET_NAME: ${{ secrets.SENTRY_DSN_SECRET_NAME }}
           SANDBOX_CLEANUP_ENABLED: ${{ vars.SANDBOX_CLEANUP_ENABLED }}
           SANDBOX_CLEANUP_PROVIDER: ${{ vars.SANDBOX_CLEANUP_PROVIDER }}
           SANDBOX_CLEANUP_SECRET_NAME: ${{ vars.SANDBOX_CLEANUP_SECRET_NAME }}
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -240,13 +240,13 @@ jobs:
       AWS_DEFAULT_REGION: ${{ vars.AWS_REGION }}
       CDK_DEFAULT_ACCOUNT: ${{ vars.DEV_ACCOUNT_ID }}
       CDK_DEFAULT_REGION: ${{ vars.AWS_REGION }}
-      DESCOPE_MANAGEMENT_KEY_SECRET_NAME: ${{ secrets.DESCOPE_MANAGEMENT_KEY_SECRET_NAME }}
+      DESCOPE_MANAGEMENT_KEY_SECRET_NAME: ${{ vars.DESCOPE_MANAGEMENT_KEY_SECRET_NAME }}
       DESCOPE_PROJECT_ID: ${{ vars.DESCOPE_PROJECT_ID }}
       DEV_ACCOUNT_ID: ${{ vars.DEV_ACCOUNT_ID }}
       OPERATION: ${{ github.event_name == 'push' && 'deploy' || inputs.operation }}
       PRODUCTION_ACCOUNT_ID: "613431292675"
       SCOPE: ${{ github.event_name == 'push' && 'core' || inputs.scope }}
-      SENTRY_DSN_SECRET_NAME: ${{ secrets.SENTRY_DSN_SECRET_NAME }}
+      SENTRY_DSN_SECRET_NAME: ${{ vars.SENTRY_DSN_SECRET_NAME }}
 
     steps:
       - name: Validate dev configuration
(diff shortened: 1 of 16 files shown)


.github/workflows/deploy.yaml
```diff
@@ -198,14 +200,32 @@ jobs:
           npm install -g aws-cdk@2.1132.0
           uv sync --frozen --python 3.12 --project infra
 
+      - name: Validate production configuration
+        env:
```
tthuwng: P1: This job does not reference `environment: prod`, so `${{ secrets.* }}` cannot read prod environment secrets. The check will always see empty values and fail production core deployments. Add the prod environment to this job, or use repository secrets intentionally.

.github/workflows/deploy.yaml
```diff
@@ -669,6 +695,22 @@ jobs:
...
```
tthuwng: P1: This validation runs for executor-only deployments too. Those deployments use `SCOPE=executor --exclusively` and do not deploy Tracker resources, but this check still requires `SENTRY_DSN_SECRET_NAME`. Executor-only production deploys will fail without an unrelated Tracker setting. Gate this validation on a core deployment.

infra/tracker_stack.py
```diff
@@ -171,10 +172,13 @@ def __init__(
 
         descope_secrets: dict[str, aws_ecs.Secret] = {}
         if auth_required.lower() == "true":
+            descope_management_key_secret_name = (
```
tthuwng: P1: Non-production stages force authentication on, so missing configuration resolves to `YourDescopeManagementKeySecret`. Release-test and direct `make deploy` paths do not run the workflow validation, then deploy a task that cannot read a real secret. Fail closed for real synthesis, and pass placeholders only from offline synthesis jobs.

infra/tracker_stack.py
```diff
@@ -152,7 +153,7 @@ def __init__(
             "DB_PASSWORD": aws_ecs.Secret.from_secrets_manager(db_credentials_secret, field="password"),
```
tthuwng, line 152: P2: An empty `SENTRY_DSN_SECRET_NAME` silently removes the `SENTRY_DSN` mapping from production. The workflow check does not protect direct or break-glass CDK deployments. Fail in the production stack when this setting is absent, or retain an intentional production default.

tthuwng (changes requested):
I found four deployment-path issues. The production core validation cannot access environment secrets, executor-only production deployments require unrelated Tracker settings, missing non-production Descope configuration uses an invalid placeholder, and direct production CDK deployments can silently remove Sentry. Please address the inline comments.

.github/workflows/deploy.yaml
```diff
@@ -205,7 +207,9 @@ jobs:
           DEPLOYMENT_NOTIFICATIONS_SLACK_CHANNEL_ID: ${{ secrets.DEPLOYMENT_NOTIFICATIONS_SLACK_CHANNEL_ID }}
           AUTH_REQUIRED: ${{ secrets.AUTH_REQUIRED }}
           BENCHMARK_CATALOG_URL: ${{ secrets.BENCHMARK_CATALOG_URL }}
+          DESCOPE_MANAGEMENT_KEY_SECRET_NAME: ${{ secrets.DESCOPE_MANAGEMENT_KEY_SECRET_NAME }}
```
tthuwng, line 210: P1: `deploy-production-core` still has no `environment: prod`, but these values use `${{ secrets.* }}` and the stack now requires them. GitHub environment secrets are unavailable unless the job references that environment, so the normal production core deployment receives empty names and fails during CDK synthesis. Add `environment: prod` to this job or use repository-level secrets intentionally.

.github/workflows/deploy.yaml
```diff
@@ -691,7 +703,7 @@ jobs:
         env:
```
tthuwng, line 703: P1: Executor-only production deploys still do not provide `SENTRY_DSN_SECRET_NAME` or `DESCOPE_MANAGEMENT_KEY_SECRET_NAME`. `infra/app.py` always constructs `TrackerStack`, and production synthesis now raises when SENTRY is absent. Thus `make deploy STAGE=prod SCOPE=executor` fails before deploying `WorkerStack`. Export the required Tracker inputs here, or make stack construction respect the selected scope.

.github/workflows/maintenance-classification.yaml
```diff
@@ -84,6 +84,8 @@ jobs:
             CDK_DEFAULT_REGION=us-east-1 \
             AWS_REGION=us-east-1 \
```
tthuwng, line 85: P1: The required `pull_request_target` check still runs the workflow definition from the `dev` base branch. That definition does not contain these offline secret variables, so the current `synthesize-head` job fails with `Authenticated deployments require DESCOPE_MANAGEMENT_KEY_SECRET_NAME` before this PR can merge. Update the base workflow first, or provide a transition that lets the required check pass before merge.

infra/constants.py
```diff
@@ -50,7 +50,7 @@
 SANDBOX_CLEANUP_LOG_GROUP_NAME = "/valkyrie/sandbox-cleanup"
 SANDBOX_CLEANUP_SCHEDULE_NAME = "valkyrie-sandbox-cleanup"
 SANDBOX_CLEANUP_DLQ_NAME = "valkyrie-sandbox-cleanup-dlq"
-SANDBOX_CLEANUP_SECRET_NAME = "AgenticHarnessSecrets"
+SANDBOX_CLEANUP_SECRET_NAME = "YourSandboxProviderSecret"
```
tthuwng, line 53: P2: `YourSandboxProviderSecret` is a documentation placeholder, not a usable production secret. When cleanup is enabled and `SANDBOX_CLEANUP_SECRET_NAME` is absent, `ExecutorStack` still imports this name and deploys a Lambda that cannot read provider credentials. Fail closed when cleanup is enabled and the setting is missing, instead of using a fake fallback.

tthuwng (changes requested):
The updated pass still finds four blockers: production core lacks the prod environment, executor-only production synthesis lacks required Tracker inputs, the pull_request_target check fails before merge because its base workflow lacks the new variables, and cleanup still has a fake secret fallback.

.github/workflows/deploy.yaml
BradenBug: These resolve today: `DESCOPE_MANAGEMENT_KEY_SECRET_NAME`, `SENTRY_DSN_SECRET_NAME`, and `SANDBOX_CLEANUP_SECRET_NAME` exist as repository-level secrets, which `${{ secrets.* }}` reads without an `environment:` key — the same mechanism as the pre-existing `secrets.DESCOPE_PROJECT_ID` / `secrets.AUTH_REQUIRED` lines on this job. Moving these inputs to Environment secrets and adding `environment: prod` to this job is #673, stacked on this PR.

.github/workflows/deploy.yaml
BradenBug: Fixed in 167ed9a3 — the executor deploy step now exports `AUTH_REQUIRED`, `DESCOPE_PROJECT_ID`, `DESCOPE_MANAGEMENT_KEY_SECRET_NAME`, and `SENTRY_DSN_SECRET_NAME`, so the unconditional TrackerStack construction during executor-scope synth gets the same inputs as the core deploy job.

.github/workflows/maintenance-classification.yaml
BradenBug: Sequencing restored: #672 (the base-workflow change alone) is reopened and merges first — it's green because it changes no stack code. Once it's on dev, this PR's classification passes on re-run, since `pull_request_target` reads the live base branch definition; no rebase needed. The prod branch's copy will need the same env lines before the promotion that carries this change.

BradenBug (approved):

tthuwng (approved):

PR #694, opened 2026-08-17
fix(tracker): finalize forced stops before provider cleanup

conjfrnk:
## Problem

EMB timeouts take the force-stop path. Provider teardown errors can then surface as obscure `Broken TTY` or Daytona failures instead of a clear timeout outcome such as “Agent didn't finish in time.” Other benchmark paths had addressed this behavior, but EMB still exposed the gap.

This PR is the short-term fix: make our database state authoritative immediately while allowing provider cleanup to finish independently. A longer-term timeout-specific error/lifecycle treatment can follow separately.

## Summary

- Force-stop finalizes Valkyrie database state immediately: active tasks become `STOPPED`, the whole run becomes `STOPPED`, and its active dispatches become `FAILED`.
- Provider cleanup is isolated from run state: after the database commit, force-stop sends Daytona deletion signals without draining or waiting for sandbox destruction.
- Provider failures are logged without changing the already-final database result.
- Graceful stop keeps its existing `STOPPING` behavior; selected-task force does not revoke a shared dispatch while other tasks remain active.

## Verification

- `uv run pytest tests/unit -q` — 575 passed
- Focused recovery/state/dispatch tests — 72 passed
- Ruff and `git diff --check` — passed
- CI `tracker-live-integration-tests` — passed
- Postgres integration tests were collected but not executed locally because Docker is unavailable.
- Local live Daytona tests were not executed because `TEST_DAYTONA_SECRET_NAME` is unavailable.

### DIFF
--- a/services/tracker/src/tracker/utils/run_control.py
+++ b/services/tracker/src/tracker/utils/run_control.py
@@ -43,12 +43,7 @@ def apply_stop_benchmark(
     """Apply the Stop state transition without committing the transaction."""
     # Stop and recovery both update the benchmark and its tasks. Lock the benchmark
     # first so every lifecycle transition uses the same lock order.
-    session.exec(
-        select(col(Benchmark.id))
-        .where(col(Benchmark.id) == benchmark_row.id)
-        .where(col(Benchmark.org_id) == org.id)
-        .with_for_update()
-    ).one()
+    fetch_benchmark_row(benchmark_row.id, session, org, for_update=True)
 
     stoppable_statuses = [TaskStatus.PENDING, TaskStatus.BUILDING, TaskStatus.EVALUATING]
     if force:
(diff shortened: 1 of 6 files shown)


services/tracker/src/tracker/utils/run_control.py
```diff
@@ -196,7 +251,14 @@ async def force_stop_sandboxes(
         .where(col(Task.status).notin_(finished_statuses))
     ).one()
 
-    if not tasks_still_running:
+    if deletion_deferred:
+        # Keep the run retryable while ownership is ambiguous. A later force stop can
+        # re-list these sandboxes after the owner exits and complete the cleanup.
+        if not tasks_still_running:
```
OrestesK: nit: have this bool first and deletion prefered once inside

services/tracker/src/tracker/utils/run_control.py
```diff
@@ -48,18 +41,37 @@ def apply_stop_benchmark(
     task_ids: list[str] | None = None,
 ) -> None:
     """Apply the Stop state transition without committing the transaction."""
+    stoppable_statuses = [TaskStatus.PENDING, TaskStatus.BUILDING, TaskStatus.EVALUATING]
```
OrestesK, line 48: i feel like this list exists in a bunch of places?

OrestesK (approved):

PR #726, opened 2026-08-20
fix: recover expired executor dispatches

conjfrnk:
## What changed
- Persist the exact task IDs assigned to each dispatch, plus claim deadline, heartbeat, lease expiry, and durable failure reason.
- Renew a database-time lease from the owning executor host and reject stale/duplicate claims through the existing dispatch authority checks.
- Reconcile expired RUNNING dispatches in Tracker with a compare-and-set expiry check, then reuse the existing exact-task failure cleanup path.
- Backfill lease/assignment state for existing dispatch rows and add focused regression coverage.

## Safety
- A fresh heartbeat wins the reconciliation race.
- Cleanup is limited to the persisted dispatch assignment and existing timestamp boundary.
- Expiry fails owned work; it never retries provider/model work or changes task state outside that dispatch.
- #2227 can query the durable fields for alerting without inferring from logs or counts.

## Validation
- 632 Tracker unit tests passed.
- Disposable PostgreSQL executor-store test passed, including claim, heartbeat, and terminalization.
- Disposable PostgreSQL migration and schema tests passed.
- Ruff check and format checks passed.

The repository-default Alembic check against localhost:5432 was unavailable because no local PostgreSQL server was running; the container-backed migration checks passed.

---

<a href="https://app.devin.ai/review/vals-ai/valkyrie/pull/726" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


services/tracker/src/tracker/executor/release_control.py
```diff
@@ -204,6 +207,9 @@ def create_executor_dispatch(
         executor_artifact_uri=release.artifact_uri,
         executor_artifact_digest=release.artifact_digest,
         executor_protocol_version=release.protocol_version,
+        created_at=created_at,
+        assigned_task_ids=list(task_ids) if task_ids is not None else None,
+        claim_deadline_at=created_at + timedelta(seconds=DEFAULT_EXECUTOR_DISPATCH_CLAIM_TIMEOUT_SECONDS),
```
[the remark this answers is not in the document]
conjfrnk: Addressed in d38d0db1: ExecutorHost claims now require claim_deadline_at > CURRENT_TIMESTAMP. Tracker reconciliation also atomically fails expired QUEUED dispatches with CLAIM_DEADLINE_EXPIRED and cleans only their persisted task assignment, so stale queue entries cannot block completion. Added unit and disposable-Postgres coverage.

services/tracker/src/tracker/executor/dispatch_control.py
```diff
@@ -239,18 +258,64 @@ def record_dispatch_failure(
...
+    ).all()
+    recovered_count = 0
+    for dispatch_id in (*expired_running_ids, *expired_queued_ids):
+        dispatch = session.get(ExecutorDispatch, dispatch_id)
+        if dispatch is None:
+            continue
+        benchmark = session.get(Benchmark, dispatch.benchmark_id)
+        if benchmark is None:
+            continue
+        is_running = dispatch.status == ExecutorDispatchStatus.RUNNING
```
JarettForzano, line 296: **Blocking:** An expired `RUNNING` dispatch can finish between the ID scan and this reload. This then treats `FINISHED` as queued, and `record_dispatch_failure()` can rewrite it to `FAILED` because its claim deadline has expired. Preserve the scanned status, skip terminal rows, and add a finish-versus-recovery test.

services/tracker/src/tracker/executor/dispatch_control.py
```diff
@@ -239,18 +258,64 @@ def record_dispatch_failure(
...
+        failure_reason = "LEASE_EXPIRED" if is_running else "CLAIM_DEADLINE_EXPIRED"
+        if record_dispatch_failure(
```
JarettForzano, line 298: **Blocking:** If the host dies after graceful stop sets the benchmark to `STOPPING`, this fails the last dispatch and its tasks, but `record_dispatch_failure()` only terminalizes `IN_PROGRESS` benchmarks. The run remains `STOPPING`, and retry/resume rejects it. When no active dispatch remains, transition `STOPPING` to `STOPPED` and cover both queued and running expiry.

services/tracker/src/executor_protocol.py
```diff
@@ -12,6 +12,13 @@
 DEFAULT_STABLE_QUEUE_NAME = "valkyrie-stable"
 DEFAULT_EXECUTOR_RELEASE_PREFIX = "releases"
 
+# A dispatch owner renews this lease from the executor host. The Tracker
+# reconciler runs every minute, so a dead host is recovered within roughly six
+# minutes while a healthy dispatch has ample time between heartbeats.
+DEFAULT_EXECUTOR_DISPATCH_CLAIM_TIMEOUT_SECONDS = 120
```
JarettForzano: **Question:** Is every valid dispatch expected to reach an ExecutorHost within 120 seconds? With expiry reconciliation, ordinary capacity saturation or an executor restart beyond this limit converts queued work into task and benchmark failures. If this is an intentional service limit, please document and monitor it. Otherwise make the deadline capacity-aware or configurable.

tests/unit/executor_host/test_supervisor.py
```diff
@@ -73,6 +74,10 @@ async def is_current(self, authority: DispatchAuthority) -> bool:
             return self.authority_results[0]
         return self.authority_results.pop(0)
 
+    async def heartbeat(self, authority: DispatchAuthority) -> bool:
```
JarettForzano, line 77: **Suggestion:** No test verifies that `run_executor_dispatch()` starts and stops the heartbeat loop. This fake records calls, but nothing asserts them; the PostgreSQL test covers only `store.heartbeat()`. Add a deterministic test that observes a renewal and verifies cleanup stops further renewals.

services/tracker/src/tracker/database/models.py
```diff
@@ -280,6 +281,11 @@ class ExecutorDispatch(SQLModel, table=True):
     created_at: datetime = Field(default_factory=lambda: datetime.now(ZoneInfo("UTC")))
     started_at: datetime | None = None
     finished_at: datetime | None = None
+    assigned_task_ids: list[str] | None = Field(default=None, sa_column=Column(JSON, nullable=True))
+    claim_deadline_at: datetime | None = None
+    heartbeat_at: datetime | None = None
+    lease_expires_at: datetime | None = None
+    failure_reason: str | None = None
```
JarettForzano: **Suggestion:** `resolve_enqueue_failure()` still changes the dispatch to `FAILED` without setting this field, and its task error uses `cause_code=None`. Operational queries cannot distinguish a known enqueue failure from an uncoded failure. Persist a stable reason such as `ENQUEUE_FAILED` on both records, or narrow the PR description if this field intentionally covers only executor and expiry failures.

JarettForzano (changes requested):
Two correctness issues need changes before merge:

- Recovery can reload a terminal dispatch and rewrite it as `FAILED` by treating every non-`RUNNING` state as queued.
- Expiry of the last dispatch can leave a graceful stop permanently `STOPPING`.

Preserve the status captured by the expiry scan, skip terminal rows, and explicitly terminalize `STOPPING` when no active dispatch remains. The heartbeat lifecycle also needs a focused test.

JarettForzano (approved):

PR #733, opened 2026-08-21
feat: add external production environment

BradenBug:
### Why

Set up external prod yo

---

This gives externally facing Valkyrie workloads their own production account and deployment lane. The existing benchmark account is called `bench` in source configuration, while its deployed stacks and resources keep their current names. Every push to the Git `prod` branch starts both deployment lanes independently, so they can run at the same time and fail separately.

AWS account IDs come from protected GitHub secrets. The existing benchmark lane keeps the `prod` GitHub environment and its current AWS role names. The new production lane uses the `prod-external` GitHub environment and separate `ValkProd*` roles and stacks.

Closes vals-ai/valkyrie-ticket-library#133

## Stack

- **[#733](https://github.com/vals-ai/Valkyrie/pull/733) ← this PR**
- [#747](https://github.com/vals-ai/Valkyrie/pull/747)

## What changed

- Adds `bench` and `prod` source stages backed by separate AWS accounts. `bench` targets the existing benchmark account; `prod` targets the new externally facing account.
- Preserves the benchmark account's current CloudFormation stack IDs, physical resource names, SSM paths, DNS name, GitHub environment, OIDC subject, deployment concurrency key, and public RDS setting.
- Creates the external production RDS instance with encrypted storage and without public network accessibility. Valkyrie services connect to it inside the VPC.
- Keeps the existing dev and bench database encryption settings unchanged. vals-ai/valkyrie-ticket-library#236 tracks their data-preserving migration.
- Adds independent core and executor jobs for the new production account to the existing production-merge workflow.
- Reads both account IDs from GitHub secrets and rejects missing, malformed, or identical account selections before requesting short-lived AWS credentials.
- Gives the new production account isolated `ValkProd*` resources, DNS, deployment roles, and workflow concurrency.
- Lets hosted CLI configuration select `bench` or `prod`. Existing `VALKYRIE_ENV=prod` automation continues to select the benchmark environment; `VALKYRIE_ENV=external` selects the new production environment.
- Keeps deployment maintenance asserted when the new production deployment fails, preventing worker admission from reopening against a partial deployment.

## How to review

Start with `.github/workflows/deploy.yaml` for the two independent deployment lanes and their account checks. Then read `infra/deployment_target.py`, `infra/stage.py`, and `infra/stage_config.py` for the boundary between source names and deployed resource identity. Finish with `src/valkyrie/cli/runtime_config.py` for CLI selection compatibility.

```mermaid
flowchart LR
    M[Push to Git prod branch] --> B[Bench deployment jobs]
    M --> P[Production deployment jobs]
    B --> BA[Existing benchmark AWS account]
    P --> PA[New externally facing AWS account]
```

## Testing

Not yet live-tested. The new production deployment requires the AWS account from vals-ai/valkyrie-ticket-library#223, the `prod-external` GitHub environment, its deployment secrets, and AWS bootstrap roles. The first production merge must confirm that both job groups start independently, that the benchmark deployment updates its existing stacks in place, and that the new Tracker responds through its production DNS name.

Design: architectural

### DIFF
--- a/.github/scripts/synthesize-worker-templates.sh
+++ b/.github/scripts/synthesize-worker-templates.sh
@@ -1,0 +1,85 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# PR classification runs this helper from the trusted base revision while
+# synthesizing candidate source with synthetic account IDs and no AWS access.
+revision="${1:?revision is required}"
+target_branch="${2:?target branch is required}"
+label="${3:?label is required}"
+artifact_directory="${4:?artifact directory is required}"
+repository_directory="${5:-.}"
+
+source_directory="$RUNNER_TEMP/source-$label"
+mkdir -p "$source_directory" "$artifact_directory"
+: > "$artifact_directory/stack-ids.txt"
+if [[ ! "$revision" =~ ^0+$ ]]; then
+  git -C "$repository_directory" archive "$revision" | tar -x -C "$source_directory"
+fi
+
+has_bench_stage=false
+if [[ ! "$revision" =~ ^0+$ ]] && python3 -c \
+  'import ast, sys; from pathlib import Path; tree = ast.parse(Path(sys.argv[1]).read_text()); matches = [node for node in tree.body if isinstance(node, ast.Assign) and any(isinstance(target, ast.Name) and target.id == "BENCH" for target in node.targets)]; raise SystemExit(len(matches) != 1 or not isinstance(matches[0].value, ast.Constant) or matches[0].value.value != "bench")' \
+  "$source_directory/infra/stage.py"; then
+  has_bench_stage=true
+fi
+
+production_account_id=222222222222
+if [[ "$has_bench_stage" == "true" ]]; then
+  production_account_id=333333333333
+fi
+
+case "$target_branch" in
+  dev)
+    targets=("dev:ValkDevWorkerStack:111111111111")
+    ;;
+  prod)
+    if [[ "$has_bench_stage" == "true" ]]; then
+      targets=(
+        "bench:WorkerStack:222222222222"
+        "prod:ValkProdWorkerStack:333333333333"
+      )
+    else
+      targets=(
+        "prod:WorkerStack:222222222222"
+        "unsupported:ValkProdWorkerStack:333333333333"
+      )
+    fi
+    ;;
+  *)
+    echo "::error::Unsupported deployment branch '$target_branch'."
+    exit 1
+    ;;
+esac
+
+for target in "${targets[@]}"; do
+  IFS=: read -r stage stack_id account_id <<< "$target"
+  printf '%s\n' "$stack_id" >> "$artifact_directory/stack-ids.txt"
+  template="$artifact_directory/$stack_id.template.json"
+  if [[ "$revision" =~ ^0+$ || "$stage" == "unsupported" ]]; then
+    printf '{"Resources": {}}\n' > "$template"
+    continue
+  fi
+
+  assembly="$RUNNER_TEMP/cdk-$label-$stage"
+  (
+    cd "$source_directory/infra"
+    BENCH_ACCOUNT_ID=222222222222 \
+    PRODUCTION_ACCOUNT_ID="$production_account_id" \
+    DEV_ACCOUNT_ID=111111111111 \
+    CDK_DEFAULT_ACCOUNT="$account_id" \
+    CDK_DEFAULT_REGION=us-east-1 \
+    AWS_REGION=us-east-1 \
+    STAGE="$stage" \
+    AWS_DEPLOYMENT_ROLE_ORG_IDS=00000000-0000-0000-0000-000000000001 \
+    AWS_TRACKER_SECRET_NAME_PREFIXES=offline-synth \
+    AWS_EXECUTOR_SECRET_NAME_PREFIXES=offline-synth \
+    BENCHMARK_CATALOG_URL=https://offline.invalid \
+    SENTRY_DSN_SECRET_NAME=offline-synth \
+    DESCOPE_MANAGEMENT_KEY_SECRET_NAME=offline-synth \
+    DESCOPE_PROJECT_ID=offline-synth \
+    CDK_CONTEXT_JSON="{\"stage\":\"$stage\"}" \
+    CDK_OUTDIR="$assembly" \
+      uv run --frozen python app.py
+  )
+  cp "$assembly/$stack_id.template.json" "$template"
+done
--- a/.github/workflows/maintenance-classification.yaml
+++ b/.github/workflows/maintenance-classification.yaml
@@ -31,6 +31,7 @@ jobs:
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
         with:
           fetch-depth: 1
+          path: trusted
           persist-credentials: false
           ref: ${{ env.BASE_SHA }}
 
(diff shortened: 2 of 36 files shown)


.github/scripts/synthesize-worker-templates.sh
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env bash
```
BradenBug, line 1: what is this file

.github/scripts/validate-production-account.sh
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env bash
```
BradenBug: what is this file

.github/workflows/maintenance-classification.yaml
```diff
@@ -284,9 +235,12 @@ jobs:
           from pathlib import Path
 
           target_branch = os.environ["TARGET_BRANCH"].removeprefix("refs/heads/")
-          stack_ids = {"dev": "ValkDevWorkerStack", "prod": "WorkerStack"}
+          stack_ids = {
+              "dev": ["ValkDevWorkerStack"],
+              "prod": ["WorkerStack", "ValkProdExternalWorkerStack"],
```
BradenBug: we can name the external prod stacks just "ValkProd..." instead of "ValkProdExternal..."

infra/README.md
```diff
@@ -13,9 +13,19 @@ AWS CDK infrastructure for the Agentic Harness benchmark platform.
...
 existing stack and retained resources in place.
 
-Production imports the existing `vals.ai` hosted zone. Development imports only
-the account-local `benchmark-tracker-dev.vals.ai` child zone. Each production
-and development Tracker stack creates and owns its ACM certificate.
+Internal production imports the existing `vals.ai` hosted zone. Development and
+external production import an account-local delegated child zone. Each Tracker
+stack creates and owns its ACM certificate in the account where it runs.
+
+The `prod` and `prod-external` stages are both production: same sizing, log
```
BradenBug: "same sizing, log retention..." - these should be parameterized. I would like to be able to scale the internal and external prod svcs independently

infra/README.md
```diff
@@ -50,17 +61,36 @@ Secrets Manager secret containing the DSN. Production also requires
 `DESCOPE_MANAGEMENT_KEY_SECRET_NAME`
 whenever `AUTH_REQUIRED` is `true`. The dev Environment holds every dev
 deployment input as an Environment secret except `AWS_REGION`, which stays a
-variable. Production reads its managed AWS inventory from the `prod`
-Environment and retains the existing repository-level deployment secrets. The
+variable. Internal production reads its managed AWS inventory from the `prod`
```
BradenBug: too much

infra/README.md
```diff
@@ -184,13 +220,133 @@ make plan STAGE=release-test SCOPE=all AWS_REGION=us-east-1 \
   DEV_ACCOUNT_ID="$DEV_ACCOUNT_ID" PROFILE=vals-dev-admin
 ```
 
+### Bootstrapping `prod-external`
```
BradenBug: too much. this should be internal doumenation

README.md
```diff
@@ -38,6 +38,17 @@ valkyrie config init
 
 This will prompt you to choose between **hosted** and **self-hosted** mode, then collect the required credentials. See [Hosted vs Self-Hosted Mode](docs/HOSTED_MODE.md) for detailed setup instructions.
 
+Vals users who were assigned to external production must select it before
+initializing configuration or running other commands:
+
+```bash
+export VALKYRIE_ENV=external
```
BradenBug: hmmm.... not a good ux

.github/scripts/synthesize-worker-templates.sh
BradenBug: This is the trusted offline synthesis helper used by deployment and PR maintenance classification. It synthesizes the internal and external WorkerStack templates for exact Git revisions with synthetic account IDs and no AWS access. The PR workflow runs the base revision's copy so candidate code cannot replace the orchestration.

src/valkyrie/cli/runtime_config.py
```diff
@@ -3,32 +3,66 @@
...
+EXTERNAL_ENVIRONMENT = "external"
 DEV_ENVIRONMENT = "dev"
 
 VALKYRIE_ENV_ENV_VAR = "VALKYRIE_ENV"
 TRACKER_SERVICE_URL_ENV_VAR = "TRACKER_SERVICE_URL"
 VALKYRIE_CONFIG_PATH_ENV_VAR = "VALKYRIE_CONFIG_PATH"
+ENVIRONMENT_CONFIG_KEY = "environment"
 
 PROD_TRACKER_URL = "https://benchmark-tracker.vals.ai"
+EXTERNAL_TRACKER_URL = "https://benchmark-tracker-external.vals.ai"
```
BradenBug: what do we think?

infra/tests/test_monitoring_stack.py
```diff
@@ -274,13 +383,14 @@ def test_tracker_transport_follows_stage_contract(self) -> None:
         tracker_templates: dict[str, assertions.Template] = {}
         for stage_name, environment in (
             (PROD, TEST_PROD_ENV),
+            (PROD_EXTERNAL, TEST_PROD_EXTERNAL_ENV),
```
jaewonjeung-vals: @BradenBug I named the aws account as dev, bench, prod. Having a separate identification (prod, external prod, dev), while it's pretty nice, could be confusing in the future. 
Note: https://github.com/vals-ai/valkyrie-ticket-library/issues/133

infra/README.md
```diff
@@ -26,16 +29,16 @@ and development Tracker stack creates and owns its ACM certificate.
...
+zone must delegate its Tracker child zone to the target account so CDK DNS
+validation can complete. Configure the protected `dev` GitHub
 Environment with the `AWS_REGION=us-east-1` variable and the `DEV_ACCOUNT_ID`,
 `AWS_DEPLOY_ROLE_ARN`, `DESCOPE_PROJECT_ID`, and
 `DESCOPE_MANAGEMENT_KEY_SECRET_NAME` secrets (the last one names an
 account-local Secrets Manager secret holding the Descope management key).
 Managed AWS execution also requires these secrets in each enabled stage's
-protected GitHub Environment. The `dev` and `prod` environments hold their own
-values:
+protected GitHub Environment. The `dev`, `prod`, and `prod-external`
```
jaewonjeung-vals, line 40: @BradenBug interesting, so on github, we're still using prod and prod-external? wondering if there's a decent technical challenge for this or if this was left out mistakenly

jaewonjeung-vals: The reason why I ask this and hopefully also becomes dev bench prod is that if we're in the "prod" context elsewhere, on github "prod" is actually bench. It could cause confusion

jaewonjeung-vals:
Codex briefly mentioned this
```
The dev merge updated bench and dev but not the new prod. e4100e2b added _TRACKER_ANALYZER_LAMBDA_PATTERNS (analysis-*) and expanded _EXECUTOR_OUTPUT_LAMBDA_PATTERNS to six final-view lambdas. BENCH_CONFIG and DEV_CONFIG got both; PROD_CONFIG still has
  executor_lambda_function_name_patterns=("vals-format-lambda",) and no tracker patterns at all. Could be correct if external prod won't run those benchmarks — but it reads like a merge miss, and it'll surface as runtime AccessDenied on lambda invoke rather than a
  deploy failure.
```
I'm not that familiar regarding that, but it's something that I did want to flag in case it results in env diverging

jaewonjeung-vals:
Codex also flagged the matter of executor_all_secret_access=True giving getsecretvalue on secret:*, but we'll handle this later once we split...

BradenBug:
> Codex also flagged the matter of executor_all_secret_access=True giving getsecretvalue on secret:*, but we'll handle this later once we split...

Yeah, this is a separate issue. It's kind of coupled with your API key work.

BradenBug:
> Codex briefly mentioned this
> 
> ```
> The dev merge updated bench and dev but not the new prod. e4100e2b added _TRACKER_ANALYZER_LAMBDA_PATTERNS (analysis-*) and expanded _EXECUTOR_OUTPUT_LAMBDA_PATTERNS to six final-view lambdas. BENCH_CONFIG and DEV_CONFIG got both; PROD_CONFIG still has
>   executor_lambda_function_name_patterns=("vals-format-lambda",) and no tracker patterns at all. Could be correct if external prod won't run those benchmarks — but it reads like a merge miss, and it'll surface as runtime AccessDenied on lambda invoke rather than a
>   deploy failure.
> ```
> 
> I'm not that familiar regarding that, but it's something that I did want to flag in case it results in env diverging

This is fine, these are lambdas for internal use.

infra/README.md
BradenBug: We'll rename this separately, it's a bit of a migration, don't want to do too much at once

jaewonjeung-vals:
Let's enable `storage_encrypted` to be True, especially for the prod account since external customers would be in that account. If they have like encrypt at rest requirement, we're doomed.

Codex let me know that we have to make it stage-by-stage case. If we set storage encrypted to True on dev or bench rn, it's gonna create a new rds instance and orphan the previous one. Maybe we can do that for dev, but definitely not for bench. We'll just consider that as tech debt

jaewonjeung-vals:
At first, I found the idea that deploying to prod deploys to both bench and prod cool, but I'm now second-guessing if that's a good idea. Bench is strictly internal, and I'm wondering in the future if there could be some features available for internal researchers to use and publish but not GA. GA probably doesn't need to get quick updates that often.

For a case like this maybe? https://github.com/vals-ai/Valkyrie/pull/733#issuecomment-5500218365

I propose either dev -> bench -> prod or dev -> bench+prod  but prod being feature-gated

jaewonjeung-vals:
Let's not do `publicly_accessible=stage.is_production`. This is not needed even in prod. This is attached to the RDS instances, and there's no reason it should be publicly accessible as raw. The RDS instance is being protected by only a single layer atm (security groups). We want defense in depth.

We can have this turned off for the initial deployment in prod, and have another pr for dev and bench switch off

jaewonjeung-vals (changes requested):
I think we're almost there!

BradenBug:
> At first, I found the idea that deploying to prod deploys to both bench and prod cool, but I'm now second-guessing if that's a good idea. Bench is strictly internal, and I'm wondering in the future if there could be some features available for internal researchers to use and publish but not GA. GA probably doesn't need to get quick updates that often.
> 
> For a case like this maybe? [#733 (comment)](https://github.com/vals-ai/Valkyrie/pull/733#issuecomment-5500218365)
> 
> I propose either dev -> bench -> prod or dev -> bench+prod but prod being feature-gated

It's a public repo. Also it's not really "GA" - we still control who has access to prod. I will defer this change to when a need reveals itself.

BradenBug:
> Let's not do `publicly_accessible=stage.is_production`. This is not needed even in prod. This is attached to the RDS instances, and there's no reason it should be publicly accessible as raw. The RDS instance is being protected by only a single layer atm (security groups). We want defense in depth.
> 
> We can have this turned off for the initial deployment in prod, and have another pr for dev and bench switch off

Sure, makes sense, will make a separate PR for making bench private. BTW, for this kind of comment, can you leave it on the code so I can reply directly and see the line you're pointing to directly?

infra/tracker_stack.py
jaewonjeung-vals, line 132: Were you able to consult your agent regarding `storage_encrypted`? remember to set this differently per stage. Having it True on bench and dev will create a new RDS instance.

BradenBug: oops, missed this one in the last pass, one sec.

BradenBug: done

.github/workflows/tracker-integration-tests.yaml
```diff
@@ -62,12 +62,29 @@ jobs:
           uv venv --python 3.12
           uv sync --group dev
 
+      - name: Validate bench account boundary
+        if: steps.rel.outputs.relevant == 'true'
+        env:
+          BENCH_ACCOUNT_ID: ${{ secrets.VALKYRIE_BENCH_ACCOUNT_ID }}
```
jaewonjeung-vals, line 68: Does this exist in secrets at the moment? I'm not sure if i can't find it or if it doesn't exist
Also `secrets.VALKYRIE_PRODUCTION_ACCOUNT_ID`
This might break bench deployment

BradenBug: no, required before merge, but does not require a revision on the PR

jaewonjeung-vals (approved):
🚢 itt

PR #741, opened 2026-08-31
feat(tracker): support repeatable completion lambdas

conjfrnk:
## Summary

A run could configure only one completion lambda, so a benchmark with its own final-view lambda (programbench, swebench, snap, terminal-bench, harvey) could never also produce the shared `vals_format/vals_format.json` artifact that export and cost recalculation read — the artifact had to be produced later by hand-invoking `vals-format-lambda`. `--lambda` is now repeatable and tracker invokes every configured lambda with the same completion payload.

```diff
-lambda_function: str | None
+lambda_functions: list[str]
```

Persisted `BenchmarkArguments` and inbound requests still accept the singular `lambda_function` (alias + string widening), so runs started before this change reconstruct, recover and finalize unchanged. The completion payload also still carries `lambda_function`, now naming the callback receiving it, so deployed lambdas reading the documented singular field are unaffected. Preflight dry-runs every configured lambda; finalization builds the payload once and invokes each lambda synchronously after the final-view upload.

One callback failure must not cost another its artifact, so each invocation is isolated and the first failure is re-raised only once every lambda has been attempted. `invoke_lambda` wraps only `ClientError`, so this catches any exception — a read timeout on the 60s completion config arrives as `ReadTimeoutError` and would otherwise skip every later callback.

## Changes

- SDK: `runs.start(lambda_functions=[...])`; `StartBenchmarkRequest`/`BenchmarkArguments` carry a list.
- CLI: `--lambda` is `multiple=True`.
- Tracker: one shared `CompletionLambdas` annotated type carries the list, alias and widening for both the inbound request and the persisted arguments, so the two shapes cannot drift; preflight validates each lambda; finalization invokes each one.
- Docs: `docs/LAMBDA_USAGE.md` payload fields, repeatability, and callback failure semantics.

## Related Issues

Motivated by downstream benchmark configs that need a legacy final-view lambda and the shared `vals-format-lambda` on the same run; those consumers are tracked separately.

## Type of Change

- [x] New feature
- [x] Docs / config

## Testing

- [ ] Manually tested

Not live-tested: this path only runs inside a deployed tracker at run finalization, and I did not start a benchmark run. `test_managed_execution_invokes_every_configured_completion_lambda` covers both preflight dry-runs, both invocations, and a payload that differs only in `lambda_function`; `test_managed_execution_invokes_later_completion_lambdas_after_a_failure` covers a later callback still running after both a `LambdaError` and a `ReadTimeoutError`; legacy singular payloads are covered by the parameterized `BenchmarkArguments` cases.

Note on the docs correction: `docs/LAMBDA_USAGE.md` claimed a failing lambda marks the run `ERROR`. It does not, and did not before this PR — finalization commits the terminal status before invoking callbacks, so `commit_benchmark_error` hits `require_in_progress` and rolls back. A failing callback is logged and reported, but the run stays `FINISHED`.

Deliberately not changed here. The ordering predates this PR and applies to every run that configures any completion lambda, so flipping a run whose tasks all succeeded to `ERROR` would change retry, monitoring and status semantics repo-wide — too broad to ride along with repeatable lambdas. This PR makes the behaviour accurate instead of aspirational, in the docs, in a comment at the `raise`, and in a test that asserts `FINISHED`.

## Checklist

- [x] Self-reviewed the diff
- [x] Docs updated if needed

Link to Devin session: https://app.devin.ai/sessions/43fd1e8242e44704bc575bedd476e654
Open in Devin Desktop: https://app.devin.ai/desktop/session/43fd1e8242e44704bc575bedd476e654?variant=devin
Requested by: @conjfrnk

---

<a href="https://app.devin.ai/review/vals-ai/valkyrie/pull/741" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=3">
    Devin Review
  </picture>
</a>

### DIFF
--- a/docs/LAMBDA_USAGE.md
+++ b/docs/LAMBDA_USAGE.md
@@ -7,7 +7,7 @@ valkyrie run start \
   --agent agents/claude_code \
   --benchmark swebench \
   --lambda my-post-benchmark-handler \
-  --lambda vals-format-lambda
+  --lambda my-result-archiver
 ```
 
 Each lambda is invoked once after all tasks finish and results are uploaded, in the order given. A lambda that fails (uncaught exception or `statusCode >= 400`) does not stop the remaining ones; the failure is logged and reported, and the first one is re-raised once every lambda has been attempted. The run status is already terminal by this point, so a failing lambda does not change it.
@@ -31,7 +31,7 @@ The tracker invokes your lambda with the full `BenchmarkArguments` plus the pers
   "task_ids": ["astropy__astropy-12907"],
   "slice_str": null,
   "lambda_function": "my-post-benchmark-handler",
-  "lambda_functions": ["my-post-benchmark-handler", "vals-format-lambda"],
+  "lambda_functions": ["my-post-benchmark-handler", "my-result-archiver"],
   "benchmark_id": "e532551e-d51b-4912-983d-47695bd24174",
   "benchmark_name": "swebench"
 }
--- a/docs/LAMBDA_USAGE.md
+++ b/docs/LAMBDA_USAGE.md
@@ -6,10 +6,11 @@ The `--lambda` flag on `run start` lets you invoke AWS Lambda functions after a 
 valkyrie run start \
   --agent agents/claude_code \
   --benchmark swebench \
-  --lambda my-post-benchmark-handler
+  --lambda my-post-benchmark-handler \
+  --lambda vals-format-lambda
 ```
 
-Each lambda is invoked once after all tasks finish and results are uploaded. If a lambda fails (uncaught exception or `statusCode >= 400`), the run will be marked as `ERROR`.
+Each lambda is invoked once after all tasks finish and results are uploaded, in the order given. A lambda that fails (uncaught exception or `statusCode >= 400`) does not stop the remaining ones; the failure is logged and reported, and the first one is re-raised once every lambda has been attempted. The run status is already terminal by this point, so a failing lambda does not change it.
 
 ## Payload
 
@@ -29,7 +30,8 @@ The tracker invokes your lambda with the full `BenchmarkArguments` plus the pers
   "concurrency": 5,
   "task_ids": ["astropy__astropy-12907"],
   "slice_str": null,
-  "lambda_functions": ["my-post-benchmark-handler"],
+  "lambda_function": "my-post-benchmark-handler",
+  "lambda_functions": ["my-post-benchmark-handler", "vals-format-lambda"],
   "benchmark_id": "e532551e-d51b-4912-983d-47695bd24174",
   "benchmark_name": "swebench"
 }
@@ -41,7 +43,8 @@ The tracker invokes your lambda with the full `BenchmarkArguments` plus the pers
 | `concurrency` | int | Concurrency level |
 | `task_ids` | list or null | Task IDs that were run (null = all) |
 | `slice_str` | string or null | Dataset slice if provided |
-| `lambda_functions` | list | Names of the lambdas invoked at completion |
+| `lambda_function` | string | Name of the lambda receiving this payload |
+| `lambda_functions` | list | Every lambda invoked at completion, in configured order |
 | `benchmark_id` | string | UUID of the completed run |
 | `benchmark_name` | string | Persisted name of the completed benchmark |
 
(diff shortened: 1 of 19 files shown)


conjfrnk:
Addressed the review findings; pushed as `fix(tracker): keep the singular payload field and attempt every completion lambda`.

**Legacy callbacks receive incompatible payloads (Devin, P1)** — real, fixed. The payload is `BenchmarkArguments.model_dump()`, so widening the field silently dropped `lambda_function`, which `docs/LAMBDA_USAGE.md` documents as part of the contract. Each invocation now carries `lambda_function` naming the callback receiving it, alongside the full `lambda_functions` list. `vals_format_lambda.py` never read the field, but the benchmark-specific lambdas are deployed out of tree and the docs promised it.

**Callback failure skips remaining lambdas (Greptile P1, Devin)** — the loop already isolated `LambdaError` when this was flagged, but that was not enough. `invoke_lambda` only wraps `ClientError`, so a read timeout on the 60s completion config (`Config(read_timeout=60, retries={"total_max_attempts": 1})`) arrives as `botocore.exceptions.ReadTimeoutError` and escaped the handler — skipping every later callback, which is exactly the case the repeatable flag exists to serve. Now isolates any exception per callback and re-raises the first once all have been attempted. Covered by a parameterized test over both failure shapes.

**Callbacks extend authority lock duration (Greptile P2)** — accurate, not taking it. Re-acquiring per callback would shorten each lock window but lets a revocation land between callbacks, so some callbacks run and the rest are silently skipped with the run deferred. A single hold keeps "every configured callback is attempted under one authority check" true, and the worst case is 120s for today's two-lambda configs against 60s before.

**One correction while in here:** the docs claimed a failing lambda marks the run `ERROR`. It does not, and did not before this PR — finalization commits the terminal status before invoking callbacks, so `commit_benchmark_error` fails `require_in_progress` and rolls back. A failing callback surfaces through the log and Sentry only. The docs now say that; changing the behaviour is a separate call worth making, since a silently-failed `vals-format-lambda` is what leaves cost per test null downstream.

Locally: 764 tracker tests (unit + local integration + alembic), 151 SDK/contract tests, ruff and basedpyright clean.

conjfrnk:
Status update — not abandoning this, but two things changed underneath it.

**The downstream consumer closed.** The benchmark-side change that would have configured lambda pairs is closed as largely superseded: the benchmarks that needed pairing are being migrated onto the shared canonical output path one at a time instead. Three benchmarks still carry only a legacy final-view lambda, so pairing is still one valid way to fix them — but migration is the other, and it needs nothing from this PR. So this currently has no committed consumer.

**The version bump is stale.** This bumps the SDK 0.2.9 → 0.2.10, but 0.2.10 was taken by #754 and dev is now at 0.2.11. It needs to become 0.2.12, and the branch is ~13 commits behind dev.

Independent of the above, one piece here stands alone: `docs/LAMBDA_USAGE.md` claimed a failing completion lambda marks the run `ERROR`, and it does not — finalization commits the terminal status before invoking callbacks, so `commit_benchmark_error` fails `require_in_progress` and rolls back. A failing callback is logged and reported only. That correction is worth keeping whatever happens to the rest.

Holding for a call on whether repeatable lambdas are still wanted given the migration direction.

conjfrnk:
Closing. Confirmed the direction is per-benchmark migration onto the shared canonical output path rather than invoking several completion lambdas per run, so repeatable `--lambda` has no future consumer. ProgramBench and CyberBench have already gone that way.

Nothing here is merged, so the tracker keeps its single `lambda_function` and no behaviour changes.

Two notes for whoever comes back to this area:

**One factual error survives in `docs/LAMBDA_USAGE.md` on dev.** It states that if a completion lambda fails "the run will be marked as `ERROR`". It does not, and did not before this PR: finalization commits the terminal status *before* invoking the callback, so `commit_benchmark_error` fails its `require_in_progress` check and rolls back. A failing completion lambda is logged and sent to Sentry, and the run stays `FINISHED`. That matters operationally — a completion lambda that silently fails leaves no artifact behind a green run — and it is still documented backwards. Worth a one-line docs fix independently of this PR.

**If anyone does revive multi-lambda:** `invoke_lambda` wraps only `ClientError` in `LambdaError`, so a read timeout on the 60s completion config (`Config(read_timeout=60, retries={"total_max_attempts": 1})`) surfaces as `botocore.exceptions.ReadTimeoutError` and escapes a `except LambdaError` handler. Any loop over several callbacks needs to isolate broadly, or the first slow lambda silently skips the rest. Also note the completion payload is `BenchmarkArguments.model_dump()`, so renaming the field changes the documented `lambda_function` key that deployed callbacks read.

Branch `cf/multi-completion-lambda` is left in place.

PR #747, opened 2026-09-02
fix: separate bench sentry environment

BradenBug:
### Why

Split sentry environments into `dev`, `bench`, and `production` instead of just `dev` and `production`

---

Bench currently reports Sentry events under the `production` environment because its broader runtime label remains `production`. This gives Sentry its own stage-specific label while leaving the runtime behavior unchanged.

Tracker and ExecutorHost receive `SENTRY_ENVIRONMENT=bench` from their ECS task definitions. Other stages keep their existing Sentry labels.

## What changed

- Adds a Sentry-specific environment label to each stage configuration.
- Passes the label to Tracker and ExecutorHost without changing their existing runtime environment.
- Extends the existing infrastructure and observability tests to cover the label.
- Updates both bench ECS task definitions. The normal CDK deployment replaces the running tasks without requiring maintenance approval.

## Testing

Not yet live-tested. The first bench deployment must confirm that new Tracker and ExecutorHost events appear under the `bench` environment in Sentry.

Design: not architectural (separates one observability label without changing service boundaries or runtime behavior)


infra/stage_config.py
```diff
@@ -104,6 +105,7 @@ class StageConfig:
 
 BENCH_CONFIG = StageConfig(
     runtime_environment="production",
+    sentry_environment=BENCH,
```
jaewonjeung-vals, line 108: Looks like a simple enough change, lgtm. I'm sure you're also thinking it, but the runtime_environment logic (how it still says "production") is tech debt that also needs to be changed downstream/upstream in the future

jaewonjeung-vals (approved):

PR #757, opened 2026-09-04
Add result previews for active runs

JarettForzano:
## Summary

Add S3 previews for active runs without changing run state. Archive the previous result before replacing the canonical result with the current snapshot.

## Changes

- Add a dedicated `/preview-results` endpoint that writes to S3 without requiring additional query flags.
- Add `valkyrie run results <run-id> --preview` and `client.runs.preview(run_id)`.
- Support task subsets through the CLI, SDK, and tracker, including `--task-ids-file` in the CLI.
- Validate requested task IDs against the run before scoring or publishing. Reject the entire request with HTTP 400 if any ID is unknown.
- Recompute the current score while keeping unfinished selected tasks as null inputs.
- Archive an existing canonical result under `benchmarks/<run-id>/archive/<UTC timestamp>/` before overwriting it.
- Skip the overwrite confirmation for previews. Preserve the confirmation for ordinary `--s3` requests.
- Invoke the configured completion callback after publishing each preview. Include the actual results bucket in preview and completion callback payloads.
- Publish the SDK changes as version `0.2.12`.

## Preview callback contract

After publishing a preview, the tracker synchronously invokes the run's configured `lambda_function`, if present. The payload contains the stored benchmark arguments plus:

```json
{
  "benchmark_id": "<run-id>",
  "benchmark_name": "<benchmark>",
  "bucket": "<results-bucket>",
  "preview": true
}
```

Each published preview triggers one callback invocation. If the callback fails, the request returns an error, but the published result and archive remain in S3.

## Usage

```text
valkyrie run results <run-id> --preview
valkyrie run results <run-id> --preview --task-ids task-a,task-b
valkyrie run results <run-id> --preview --task-ids-file task-ids.txt
```

```python
preview = await client.runs.preview(run_id)
subset_preview = await client.runs.preview(run_id, task_ids=["task-a", "task-b"])
```

Each request writes the tracker result to:

```text
benchmarks/<run-id>/<benchmark>.json
```

If that object already exists, the tracker first copies it to:

```text
benchmarks/<run-id>/archive/<UTC timestamp>/<benchmark>.json
```

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactor
- [x] Docs / config

## Testing

Checked items passed. Unchecked items are pending, failed, or skipped as noted.

CI for current head `c67cacc3` is running. Previous CI at `50038b3a`:

- [x] Run tracker unit and local-integration tests — **825 passed**, 90.97% coverage.
- [x] Run CLI tests — **232 passed**, 85.91% coverage.
- [x] Run documentation-reference tests — **14 passed**.
- [x] Run Ruff and basedpyright checks.
- [ ] Pass CLI startup smoke — failed importing `AsyncDaytona` from `daytona`.
- [ ] Pass SDK packaging — version `0.2.11` does not exceed the base branch's `0.2.11`.
- [ ] Run SDK compatibility tests — skipped after packaging failed.
- [ ] Run tracker live-integration tests — test steps were skipped despite a green job status.

Local verification:

- [x] Test the SDK version bump to `0.2.12` (`c67cacc3`) — version and lock checks passed; **156 SDK/contract tests passed**; wheel/sdist validation and isolated installation/namespace checks passed. Python 3.13 and 3.14 wheel import checks passed.
- [x] Run task-ID regression coverage — **68 tracker API tests passed**, including unknown-only/mixed rejection and valid unfinished tasks.
- [x] Re-test unknown-only and mixed task-ID selections live at `50038b3a` — both rejected; S3 objects and run state stayed unchanged.
- [x] Verify byte-identical archives, repeated callback invocation, overwrite refusal, and previews without a callback — tested at `6bfdbd98` using errored tasks.
- [x] Test the callback bucket fix (`0c19765b`) — **87 tests passed**; Ruff, formatting, and basedpyright passed; independent review found no blockers.
- [x] Preview two scored GPT-5 tasks plus one unfinished task — raw results contain two scored entries; formatted output contains all three selected tasks.
- [x] Verify formatted output after each preview using the local bucket fix — all four versions contain the selected task IDs. Lambda logs correlate by timestamp, not run ID.
- [x] Verify repeated two-task previews through inline IDs and an IDs file — both preserve the same task selection and score.
- [x] Verify all three archives match their preceding canonical versions byte-for-byte. An orphaned test process caused one extra preview; actual sequence was three tasks, three tasks, two tasks, two tasks.
- [x] Verify automatic completion replaces the preview — after externally deleting the remaining Java sandbox, the run finished with five scored tasks and one error. Canonical results and formatted output updated automatically; formatted output includes all six tasks; archives stayed at three. Read-only CLI retrieval returned final score **2.272826**. All-six-success completion was not exercised.
- [x] Verify downstream export consumes the preview — export passed; canonical and formatted hashes stayed unchanged throughout. The existing exporter zero-fills omitted tasks, so exported accuracy is not subset-only.

## Checklist

- [x] Self-reviewed the diff
- [x] Docs updated


services/tracker/main.py
```diff
@@ -812,56 +820,88 @@ async def analyze_benchmark(
...
     session: Session = Depends(get_session),
     org: Org = Depends(get_current_org),
 ) -> RetrieveResultsResponse:
-    """
-    Retrieve the results of a benchmark by its id. When task_ids is non-empty, the final view is
-    filtered to that subset and the final score is recomputed over the subset; the persisted
-    FinalEvaluation / per-task rows are left untouched.
+    """Retrieve a benchmark's current, subset, or versioned preview results."""
+    if preview and not s3:
+        raise HTTPException(status_code=400, detail="Preview results must be saved to S3")
```
JarettForzano: By default we should enable s3 if preview is used

services/tracker/main.py
```diff
@@ -812,56 +820,88 @@ async def analyze_benchmark(
...
-    curl -X GET http://<endpoint>/retrieve-results?benchmark_id=<uuid>&s3=false
-    curl -X GET 'http://<endpoint>/retrieve-results?benchmark_id=<uuid>&task_ids=task_1&task_ids=task_2'
-    """
     benchmark_row = assert_org(
         session.get(Benchmark, benchmark_id, options=[joinedload(Benchmark.final_evaluation)]),
         org,
     )
+    lambda_function = benchmark_row.arguments.lambda_function
+    if preview and not lambda_function:
+        raise HTTPException(status_code=400, detail="Preview results require a configured Lambda function")
```
JarettForzano: Preview should not require a lambda, since we can want partial final results computed but not use a lambda on them

JarettForzano: Actually lambda output location should not be coupled with the preview location. Revert having lambda output in preview folder.

services/tracker/main.py
```diff
@@ -812,56 +820,88 @@ async def analyze_benchmark(
...
 ) -> RetrieveResultsResponse:
-    """
-    Retrieve the results of a benchmark by its id. When task_ids is non-empty, the final view is
-    filtered to that subset and the final score is recomputed over the subset; the persisted
-    FinalEvaluation / per-task rows are left untouched.
+    """Retrieve a benchmark's current, subset, or versioned preview results."""
+    if preview and not s3:
+        raise HTTPException(status_code=400, detail="Preview results must be saved to S3")
+    if preview and task_ids:
+        raise HTTPException(status_code=400, detail="Preview results cannot be limited to task IDs")
```
JarettForzano: Why not? compute the final score based off a subset of tasks defined.

services/tracker/main.py
```diff
@@ -890,18 +930,60 @@ def _filter_task_map(task_map: dict[str, Any] | None) -> dict[str, Any] | None:
         )
 
     if s3:
-        s3_key = await upload_final_view(benchmark_row, final_view, aws_runtime)
+        preview_version: int | None = None
+        preview_prefix: str | None = None
```
JarettForzano: A lot below here should be put in different methods / broken up

services/tracker/main.py
```diff
@@ -812,99 +820,165 @@ async def analyze_benchmark(
...
+            "preview": True,
+        }
+    )
+    await asyncio.to_thread(
+        invoke_lambda,
+        aws_runtime.clients,
+        lambda_function,
+        lambda_payload,
+        config=_COMPLETION_CALLBACK_CONFIG,
+    )
```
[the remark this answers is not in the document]
JarettForzano: Not applicable to the current API contract. `Runs.preview()` returns `S3UploadResultsResponse`, containing the tracker result's `s3_url`, `presigned_url`, `console_url`, and `expires_in`. Callback-artifact locations are not part of that response. The configured callback writes its own outputs, which downstream export reads separately. Live verification confirmed formatted preview output and successful downstream ingestion. Resolving this contract mismatch; this does not resolve the separate callback-input or concurrent-overwrite concerns.

services/tracker/main.py
```diff
@@ -812,99 +820,165 @@ async def analyze_benchmark(
...
+        return
+
+    lambda_payload = benchmark_row.arguments.model_dump()
+    lambda_payload.update(
+        {
+            "benchmark_id": str(benchmark_row.id),
+            "benchmark_name": benchmark_row.name,
+            "preview": True,
+        }
+    )
```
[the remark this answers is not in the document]
JarettForzano: Fixed in 0c19765b. Both preview and completion callback payloads now include `bucket` from the same AWS runtime used to upload the result. The canonical input key is identified by `benchmark_id` and `benchmark_name`: `benchmarks/<run-id>/<benchmark>.json`; `places_where_result_data_is` is not part of the current contract. Regression coverage passed (87 affected tests), and live previews plus automatic completion produced the expected formatted S3 outputs using the configured bucket.

conjfrnk (approved):

PR #797, opened 2026-09-15
Compose runtime services and persist AWS resource settings

JarettForzano:
## Summary

Tracker API operations and benchmark execution now compose storage, secrets, logging, and sandbox services through a shared runtime interface. New runs persist resolved AWS resource locations so retries, resumes, and artifact reads retain them when deployment defaults change. This provides the factory boundary for future local runtime support; execution remains AWS-only.

## Changes

- Compose AWS adapters from resolved resources and credentials, preserving managed and caller-credential execution.
- Inject runtime services into agent storage, log routes, and benchmark execution.
- Scope sandbox providers to execution contexts with cancellation-safe cleanup and drain task log writes before shutdown.
- Share provider-secret parsing and final-result uploads across callers.
- Add `environment="aws"` and optional resource `properties` to run requests and SDK models; preserve empty legacy log-group prefixes.
- Regenerate OpenAPI and document the coordinated Tracker/executor rollout.

Rollout requires stopping and draining runs, then deploying Tracker and activating a compatible executor before reopening admission. Older executors reject the new argument fields.

## Type of Change

- [x] Bug fix
- [x] New feature
- [x] Breaking change
- [x] Refactor
- [x] Docs / config

## Testing

- [x] 878 Tracker unit tests passed on the final source tree
- [x] 25 local PostgreSQL integration tests passed on the final source tree
- [x] 236 SDK/contract tests passed before the final runtime lifecycle simplification
- [x] Lint and formatting passed; no new local type-check diagnostics
- [x] Combined branch source tree matches the tested stack tip exactly
- [x] Live AWS/provider tests rerun for the final changes

## Checklist

- [x] Self-reviewed the diff
- [x] Docs updated if needed


services/tracker/src/tracker/aws/runtime.py
```diff
@@ -36,3 +41,48 @@ def from_harness_config(cls, harness_config: HarnessConfig) -> AWSRuntime:
...
+    ) -> AsyncGenerator[RuntimeServices]:
+        """Compose existing AWS adapters without resolving credentials again."""
+        from tracker.aws.cloudwatch_logs import (
+            CloudWatchBenchmarkLogLocations,
+            CloudWatchBenchmarkLogSink,
+            CloudWatchLogProvider,
+        )
+        from tracker.aws.s3 import S3ArtifactLocations, S3ObjectStore
+        from tracker.aws.secrets import SecretsManagerStore
+        from tracker.runtime.services import RuntimeServices
```
JarettForzano: Remove out of the class

services/tracker/src/tracker/runtime/lifecycle.py
JarettForzano, line 1: Is this only used in one spot?

services/tracker/src/tracker/runtime/secrets.py
JarettForzano, line 1: These extra methods seem weird, what are they for if they are only called in one spot? When I say push down I mean into a runtime class if its used inside of the same transaction

services/tracker/src/tracker/aws/runtime.py
JarettForzano: Moved the AWS adapter imports out of the method to module scope in aws/services.py. CloudRuntimeConfig now lives there, separate from AWSRuntime. Formatting and type checks pass.

services/tracker/src/tracker/runtime/lifecycle.py
JarettForzano: It has two callers: RuntimeServices.close and TaskLogBuffer.close. Both need the same cancellation-safe cleanup, so the shared helper remains. Runtime and task cleanup tests pass.

services/tracker/src/tracker/runtime/secrets.py
JarettForzano: Provider-secret loading and parsing now live on RuntimeServices._load_sandbox_provider_config. Removed the extra helpers from runtime/secrets.py and kept existing shared resource helpers in utils/resources.py. Runtime and secret tests pass.

services/tracker/src/tracker/runtime/services.py
```diff
@@ -0,0 +1,52 @@
...
+        config = await self.get_sandbox_provider_config()
+        if self._sandbox_provider is None:
+            self._sandbox_provider = config.create_provider()
+        return self._sandbox_provider
+
+    async def close(self) -> None:
+        """Release any provider created during this operation."""
+        try:
+            if self._sandbox_provider is not None:
+                await self._sandbox_provider.close()
```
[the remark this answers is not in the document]
JarettForzano: Fixed in the current runtime lifecycle. close() marks the runtime closed before waiting on the configuration lock, and configuration loading rechecks that state after the secret lookup. Provider shutdown happens outside the lock. The regression tests cover shutdown during loading and repeated cancellation; those tests passed in the full Tracker suite.

BradenBug (approved):

PR #799, opened 2026-09-15
Use runtime services during benchmark execution

JarettForzano:
## Summary

Executors now open one runtime for their execution lifetime. Task cleanup waits for pending log writes before runtime resources close.

## Changes

- Pass composed storage, secrets, and log services into tasks.
- Resolve sandbox access lazily and close resources after finalization, including failure paths.
- Preserve managed and caller-credential execution, queue recovery, and provider selection.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor
- [ ] Docs / config

## Testing

- [x] Manually tested

## Checklist

- [x] Self-reviewed the diff
- [x] Docs updated if needed

No user-facing configuration changes in this layer.

---

<a href="https://app.devin.ai/review/vals-ai/valkyrie/pull/799" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=3">
    Devin Review
  </picture>
</a>


services/tracker/src/tracker/utils/run_orchestration.py
```diff
@@ -753,33 +748,36 @@ def record_queued_cancellation(owned_attempts: dict[UUID, datetime]) -> None:
...
+        runtime = await runtime_stack.enter_async_context(
+            CloudRuntimeConfig(properties=aws_runtime.resources).create_runtime(
+                clients=aws_runtime.clients,
+                sandbox_provider=start_benchmark_request.sandbox_provider,
+                sandbox_provider_secret_name=start_benchmark_request.get_sandbox_provider_secret_name(),
             )
+        )
+        if execution.aws_managed:
+            sandbox_provider_config = await _preflight_managed_aws(execution, aws_runtime, runtime)
+        else:
```
JarettForzano: Is there a way to push aws logic down into the aws class and just use the factory so its really just `sandbox_provider_config = await runtime.get_sandbox_provider_config()` being called?

services/tracker/src/tracker/utils/run_orchestration.py
```diff
@@ -753,33 +748,36 @@ def record_queued_cancellation(owned_attempts: dict[UUID, datetime]) -> None:
...
             notifier = SlackNotifier(
                 secret_name=start_benchmark_request.webhook_secret_name,
-                secret_store=SecretsManagerStore(aws_runtime.clients),
+                secret_store=runtime.secrets,
                 intervals=start_benchmark_request.webhook_intervals,
             )
 
         if not execution.aws_managed:
-            CloudWatchBenchmarkLogSink(aws_runtime.clients, aws_runtime.resources.log_group).create_benchmark(
+            await asyncio.to_thread(
```
JarettForzano: why did we start added `await asyncio.to_thread` everywhere?

services/tracker/src/tracker/utils/run_orchestration.py
```diff
@@ -1060,38 +1058,40 @@ def record_queued_cancellation(owned_attempts: dict[UUID, datetime]) -> None:
...
-                    notification_context = NotificationContext.from_benchmark(benchmark_row, notification_session, org)
-                    final_score = benchmark_row.final_evaluation.final_score if benchmark_row.final_evaluation else None
-                    notification_status = benchmark_row.status
-                    notification_error_message = benchmark_row.error_message
-                    await notifier.send_terminal_notification(
-                        notification_context,
-                        status=notification_status,
-                        final_score=final_score,
-                        error_message=notification_error_message,
+        async with runtime_stack:
```
JarettForzano: No depends or way to handle this in a yield block?

services/tracker/src/tracker/utils/task_execution.py
```diff
@@ -448,8 +446,20 @@ def buffer_logs(
...
+    future = loop.run_in_executor(None, log_sink.write, stream_key, message)
+    if pending_writes is not None:
+        pending_writes.add(future)
+
+        def write_finished(completed: asyncio.Future[None]) -> None:
+            pending_writes.discard(completed)
+            if not completed.cancelled() and (error := completed.exception()) is not None:
+                logger.error("Task log write failed", exc_info=(type(error), error, error.__traceback__))
+
+        future.add_done_callback(write_finished)
```
JarettForzano: Is this writing logic that can be pushed down?

services/tracker/src/tracker/types.py
```diff
@@ -104,6 +104,12 @@ def validate_name(cls, value: str) -> str:
     def validate_custom_service(cls, value: str | None) -> str | None:
         return validate_service_url_syntax(value) if value is not None else None
 
+    def get_sandbox_provider_secret_name(self) -> str | None:
+        """Resolve the provider reference from legacy or managed configuration."""
+        if self.harness_config is not None and self.harness_config.sandbox_provider_secret_name:
```
JarettForzano, line 110: This can become a property that is called instead of a boolean for ease of use

services/tracker/tests/unit/utils/task_execution_support.py
```diff
@@ -189,17 +189,20 @@ async def run_process_task(
...
-        org=TEST_ORG,
-        sandbox_provider_config=sandbox_provider_config,
-        sandbox_provider=benchmark_service.get_sandbox_provider(sandbox_provider_config),
-        creation_semaphore=Semaphore(1),
-        queue_context=queue_context,
-        authority=authority,
-    )
+    async with CloudRuntimeConfig(properties=aws_runtime.resources).create_runtime(
+        clients=aws_runtime.clients,
+    ) as runtime:
```
JarettForzano: Can we yield the runtime here?

services/tracker/src/tracker/aws/services.py
```diff
@@ -1,25 +1,55 @@
...
+        await to_thread(
+            self.logs.create_benchmark,
+            str(benchmark_id),
+            retention_days=self.aws_runtime.resources.log_retention_days,
+        )
+        if request.harness_config is not None:
+            return
+
+        await self.get_sandbox_provider_config()
+        await self.resolve_secrets(request.contract.secrets)
```
JarettForzano: This section can be improved, we just have a bunch of methods we spin off async without returning anything.

services/tracker/src/tracker/utils/run_orchestration.py
```diff
@@ -751,39 +739,23 @@ def record_queued_cancellation(owned_attempts: dict[UUID, datetime]) -> None:
...
-        else:
-            harness_config = cast(HarnessConfig, start_benchmark_request.harness_config)
-            aws_runtime = AWSRuntime.from_harness_config(harness_config)
-            sandbox_provider_config = fetch_sandbox_provider_config(
-                harness_config.sandbox_provider_secret_name,
-                SecretsManagerStore(aws_runtime.clients),
-                start_benchmark_request.sandbox_provider,
-            )
+        runtime = await runtime_stack.enter_async_context(
+            CloudRuntimeConfig.create_execution_runtime(start_benchmark_request, org.id, benchmark_id)
```
JarettForzano: Is it not possible to use a depends here? All the dependencies to make the runtime config come from the input of the method and therefore we should be able to get runtime on the fly

services/tracker/src/tracker/utils/task_execution.py
```diff
@@ -1416,12 +1371,11 @@ def sandbox_context() -> AbstractAsyncContextManager[Sandbox]:
...
-            await evaluation_lock.__aexit__(None, None, None)
-        flush_task.cancel()
-        with suppress(asyncio.CancelledError):
-            await flush_task
-        buffer_logs(log_queue, stream_key, log_sink, force_flush=True)
+        try:
+            if evaluation_lock is not None:
+                await evaluation_lock.__aexit__(None, None, None)
+        finally:
+            await task_logs.close()
```
JarettForzano: Should be able to cleanup itself instead of force this

services/tracker/src/tracker/utils/run_orchestration.py
JarettForzano: AWS construction and preflight now live in CloudRuntimeConfig.create_execution_runtime and CloudRuntimeServices.prepare_execution. Execution receives the runtime through Depends and calls runtime.get_sandbox_provider_config(). Managed execution and orchestration tests pass.

services/tracker/src/tracker/utils/run_orchestration.py
JarettForzano: Consolidated blocking AWS preflight into one synchronous prepare_execution method, called once through to_thread at the factory boundary. The worker thread keeps synchronous SDK calls off the event loop. Managed execution tests pass.

services/tracker/src/tracker/utils/run_orchestration.py
JarettForzano: Added get_execution_runtime as a yielding async context manager consumed through Depends. The outer AsyncExitStack keeps the runtime alive through finalization and closes it even if benchmark-client shutdown fails. Cleanup regression tests pass.

services/tracker/src/tracker/utils/task_execution.py
JarettForzano: Moved log buffering, periodic flushing, and final draining into TaskLogBuffer. Task execution now writes through that object and scopes it with async with. Task execution and log cleanup tests pass.

services/tracker/tests/unit/utils/task_execution_support.py
JarettForzano: The runtime_services fixture now yields the runtime from CloudRuntimeConfig.create_runtime. Task helpers accept that fixture, so fixture teardown closes the runtime. The affected task tests pass.

services/tracker/src/tracker/aws/services.py
JarettForzano: prepare_execution is now synchronous and groups log setup, secret checks, and Lambda preflight in one worker-thread call from the factory. Removed the individual async wrappers. Managed execution tests pass.

services/tracker/src/tracker/utils/run_orchestration.py
JarettForzano: Added get_execution_runtime with Depends for the validated request, benchmark, and organization. DependencyGraph resolves it for _execute_benchmark and scopes its cleanup. Startup failures still reach the existing error handling. Orchestration tests pass.

services/tracker/src/tracker/utils/task_execution.py
JarettForzano: TaskLogBuffer now implements an async context manager. Each task attempt uses async with, so the buffer drains on exit without a manual close in the task finally block. Retry and cleanup tests pass.

services/tracker/src/tracker/aws/services.py
```diff
@@ -31,12 +62,16 @@ async def create_runtime(
...
-    ) -> AsyncGenerator[RuntimeServices]:
+    ) -> AsyncGenerator[CloudRuntimeServices]:
         """Compose existing AWS adapters without resolving credentials again."""
         runtime = AWSRuntime(resources=self.properties, clients=clients)
         secrets = SecretsManagerStore(clients)
 
-        services = RuntimeServices(
+        services_type = (
+            ManagedCloudRuntimeServices if isinstance(clients, DefaultChainAWSClientProvider) else CloudRuntimeServices
+        )
```
JarettForzano: Does it even matter which object it is since they are both the same? we should just assume the factory produces `RuntimeServices` imo

services/tracker/src/tracker/utils/run_orchestration.py
JarettForzano, line 1: Why was there such a huge comment in here?

services/tracker/src/tracker/aws/services.py
JarettForzano: Both factory creation methods now yield RuntimeServices. Orchestration uses the common interface for result storage and completion callbacks, with no access to aws_runtime. The credential-source discriminator selects the concrete implementation inside the factory. Focused runtime tests and the full Tracker suite pass.

services/tracker/src/tracker/utils/run_orchestration.py
JarettForzano: The large diff came from extracting the execution body and passing shared finalization state. Removed that split and restored one guarded flow while retaining dependency resolution and cleanup. The current orchestration diff is 143 changed lines, down from 573. Cleanup and PostgreSQL finalization-lock tests pass.

services/tracker/src/tracker/aws/services.py
```diff
@@ -52,3 +107,51 @@ async def create_runtime(
...
+_RUNTIME_CONFIG_ADAPTER: TypeAdapter[RuntimeConfig] = TypeAdapter(RuntimeConfig)
+
+
+class CloudRuntimeFactory:
+    """Select configuration and open AWS services for an operation."""
+
+    @staticmethod
+    def from_aws_runtime(runtime: AWSRuntime) -> RuntimeConfig:
+        return _RUNTIME_CONFIG_ADAPTER.validate_python(
+            {"credential_source": runtime.clients.credential_source, "properties": runtime.resources}
```
JarettForzano, line 132: The Access key vs Managed runtime configs were kinda confusing. Can we deprecate Access key workflows since we don't use them internally?

PR #806, opened 2026-09-17
Add filesystem runtime adapters for local execution

JarettForzano:
## Summary

Add local-execution adapters for the async runtime interfaces from #819: filesystem artifacts, per-run JSONL logs and in-memory credential resolution. #807 and #808 add release loading and runtime selection.

Stack: dev → #819 → #822 → **#806** → #807 → #808.

## Changes

- Publish complete files atomically through a staging file. `copy` uses the same publish step for frozen agent copies.
- With `overwrite=False`, publish through a hard link. An existing key raises `FileExistsError`, matching S3 conditional uploads.
- `put_stream` checks `should_continue` once before publishing. A revoked upload leaves the previous file in place.
- `local_path` rejects absolute keys, traversal, symlink escapes and the reserved `.valkyrie` directory.
- On cancellation, wait for in-flight filesystem work before removing the staging file.
- Support concurrent log appends, filtering, pagination and previous attempts.
- Resolve declared credential references in memory. `InMemorySecretStore` raises `SecretsError` naming any missing keys.
- Move `task_log_stream_name` to `tracker.runtime.logs`, so local logs share CloudWatch stream naming. Remove the single-implementation `ObjectReadSession` protocol. `finish_cleanup` now returns the task's result.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactor
- [ ] Docs / config

## Testing

- [x] Manually tested

Local Docker runs at the #808 head exercised these adapters through a finished run, forced and normal stops, a lost sandbox, crash cleanup and retry. An AWS-mode run with Daytona also finished with logs and S3 artifacts.

## Checklist

- [x] Self-reviewed the diff
- [x] Docs updated if needed


services/tracker/src/tracker/local/storage.py
JarettForzano, line 1: Too much extra logic

services/tracker/src/tracker/local/secrets.py
```diff
@@ -0,0 +1,30 @@
...
+            self._values.setdefault(reference, {})[name] = values[name]
+
+    def get(self, name: str) -> SecretValue:
+        try:
+            return self._values[name]
+        except KeyError:
+            raise SecretsError(f"Local execution has no values for secret reference '{name}'") from None
+
+    async def get_async(self, name: str) -> SecretValue:
+        return self.get(name)
```
JarettForzano: I don't like how we have two different gets

services/tracker/src/tracker/local/storage.py
```diff
@@ -0,0 +1,150 @@
...
+    root = root.resolve()
+    path = (root / key).resolve()
+    if not path.is_relative_to(root) or (path == root and not prefix):
+        raise ValueError("Local artifact path escapes its storage directory")
+    return path
+
+
+async def _io(operation: Callable[P, T], *args: P.args, **kwargs: P.kwargs) -> T:
+    # Do not close or unlink a temporary file while a cancelled thread still uses it.
+    return await finish_cleanup(asyncio.create_task(asyncio.to_thread(operation, *args, **kwargs)))
```
JarettForzano, line 36: I think there is a simpler approach to this

services/tracker/src/tracker/local/logs.py
```diff
@@ -0,0 +1,149 @@
...
+            raise ValueError("Local log root must be absolute")
+        self.root = root.resolve()
+
+    def _path(self, benchmark_id: str) -> Path:
+        return local_path(self.root, f"{benchmark_id}/logs.jsonl")
+
+    async def create_benchmark(self, benchmark_id: str, *, retention_days: int) -> None:
+        path = self._path(benchmark_id)
+        await asyncio.to_thread(path.parent.mkdir, parents=True, exist_ok=True)
+        await asyncio.to_thread(path.touch, exist_ok=True)
```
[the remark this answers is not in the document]
BradenBug: That's... fine.

services/tracker/src/tracker/local/storage.py
```diff
@@ -0,0 +1,156 @@
...
+P = ParamSpec("P")
+T = TypeVar("T")
+
+
+def local_path(root: Path, key: str, *, prefix: bool = False) -> Path:
+    """Resolve a store key without allowing traversal or escaped symlinks."""
+    parts = PurePosixPath(key).parts
+    if (not key and not prefix) or key.startswith("/") or ".." in parts or "\x00" in key:
+        raise ValueError("Local artifact keys must be relative paths without traversal")
+    if parts and parts[0] == ".valkyrie":
```
BradenBug: suggestion: The reserved `.valkyrie` check here, the staging path at line 52 and the listing filter at line 136 all exist only because staging lives inside the key space. If staging sat beside the object root on the same filesystem, rename and link would stay atomic and all three would go away.

BradenBug (approved):

PR #807, opened 2026-09-17
Load and initialize local executor releases

JarettForzano:
## Summary

Let ExecutorHost load verified executor artifacts from a local release directory. Releases initialize through the existing database lifecycle. Rerunning the initializer with the same build keeps the active release.

Stack: dev → #819 → #822 → #806 → **#807** → #808.

## Changes

- Accept `file:///` artifact URIs in release validation.
- When `EXECUTOR_RELEASE_ROOT` is set, ExecutorHost reads artifacts only inside that directory. Otherwise it keeps the hosted S3 path.
- Verify a staged artifact before publishing it atomically under its digest. Keep prior artifact paths for saved dispatches.
- Run local artifacts from their published paths without a cache copy. Hosted S3 artifacts still use the cache.
- The initializer reuses shared admission locking and maintenance rejection.
- Write dispatch payloads to the system temporary directory instead of the cache directory. Simplify ExecutorHost S3 download staging and cancellation waits.
- Report the native OS in build metadata.

## New Commands

`python -m tracker.local.releases --manifest <manifest.json> --release-root <absolute-directory>` initializes a release. A changed build becomes the active release.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactor
- [x] Docs / config

## Testing

- [x] Manually tested

A native ARM64 PEX ran directly from its published release path. Replacing the release kept the previous executor artifact readable. Runs at the #808 head finished with score 100 in local Docker mode and in AWS mode with Daytona.

## Checklist

- [x] Self-reviewed the diff
- [x] Docs updated if needed


services/tracker/src/tracker/aws/executor_artifacts.py
```diff
@@ -14,9 +17,31 @@ def get_object(self, *, Bucket: str, Key: str) -> Mapping[str, object]: ...
...
+        expected_bucket: str | None = None,
+        expected_prefix: str | None = None,
+    ) -> None:
         self._client = client
-
-    def open(self, bucket: str, key: str) -> AbstractContextManager[BinaryIO]:
-        response = self._client.get_object(Bucket=bucket, Key=key)
+        self._expected_bucket = expected_bucket
+        self._expected_prefix = expected_prefix
+
```
JarettForzano: why are we modifying S3 here if this is for local env?

BradenBug:
question: Stepping back on the design. Local setup today means building an ARM64 PEX, registering a release, and running ExecutorHost as a separate worker (`docs/contributing/local-development.mdx`). That is about 12 manual steps, and every executor code change needs a rebuild and re-register. #808 keeps this flow. Who is the intended local user, and does local mode need executor releases at all? Could local ExecutorHost run the executor from the checkout, or could `tracker.serve --config` register a source release at startup? If so, most of this PR's release machinery and the S3-path changes go away.

PR #808, opened 2026-09-17
Run local Docker benchmarks through server-configured Tracker

JarettForzano:
## Summary

Run benchmarks locally through the existing Tracker and executor pipeline. `tracker.serve --config` selects Docker and filesystem storage. CLI and SDK clients need only the new `tracker_url` config key.

Stack: dev → #819 → #822 → #806 → #807 → **#808**.

## Changes

- Add `--config` to `tracker.serve`. The file sets `data_root` and an optional `secrets_file`. In local mode, Tracker binds to 127.0.0.1 and runs one worker.
- Use one nested `aws` config, with optional `aws.credentials`, in the SDK and CLI. The CLI reuses SDK validation, headers and harness construction.
- `ValkyrieConfig.from_yaml` rejects legacy top-level keys and points to `valkyrie config init`, which migrates them.
- Serve local agent and artifact files through the existing `download-url` routes with `download=true`. Downloads authenticate only on the configured Tracker origin.
- Store local artifacts and logs on disk. Run arguments keep the local resource paths for resume. The PR adds no database migration.
- Read contract-declared agent credentials from `secrets_file` at dispatch time.
- Support image-based Docker tasks. An AWS-mode Tracker rejects `sandbox_provider="docker"` with a 400.
- Local runs return 400 for analysis, S3 results export, Lambda callbacks, webhooks and service-auth secret references.
- Parse SDK run arguments without an `environment` field as AWS.
- Stop sending local variables to Sentry in all deployments.
- Document local benchmark-service auth (`AUTH_DISABLED=true`) and container cleanup after an executor crash.

## Configuration migration

Client YAML must use the nested format. `valkyrie config init` moves the flat AWS fields under `aws` and `aws.credentials`. It also moves `DAYTONA_SECRET_NAME` to `sandbox_providers.daytona`. Access keys already directly under `aws` must move under `aws.credentials` by hand.

Omit `aws.credentials` for managed AWS runs. Client-side AWS operations then use the default credential chain. Omit `aws` when the client needs no AWS resources.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Breaking change
- [x] Refactor
- [x] Docs / config

## Testing

- [x] Manually tested

Local Docker runs finished with score 100. They covered forced and normal stops, a lost sandbox, crash cleanup with retry, and legacy-config errors. An AWS-mode run on the dev account with Daytona finished with score 100. A forced stop deleted both Daytona sandboxes.

## Checklist

- [x] Self-reviewed the diff
- [x] Docs updated if needed


packages/valkyrie-sdk/src/valkyrie/sdk/config.py
```diff
@@ -97,49 +120,50 @@ def from_yaml(cls: type[ConfigT], path: str | Path = DEFAULT_CONFIG_PATH) -> Con
...
+            for field in AWSConfig.model_fields.values()
+            if field.alias is not None and field.alias in values
+        }
+        if any(
+            aws_values.get(key) is not None
+            for key in ("AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY", "AWS_SESSION_TOKEN")
+        ):
+            if "aws" in values:
+                raise ValkyrieConfigError("Configure either nested aws or flat AWS fields, not both")
+            values["aws"] = aws_values
```
JarettForzano: This is not a great addition

services/tracker/src/tracker/serve.py
```diff
@@ -4,15 +4,26 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--config", type=Path, help="Local execution resource configuration")
+    args = parser.parse_args()
+    if args.config is not None:
+        config.configure(args.config)
+    is_local = config.resources is not None
     uvicorn.run(
         "main:app",
-        host="0.0.0.0",
+        host="127.0.0.1" if is_local else "0.0.0.0",
```
BradenBug: suggestion: Local mode runs without auth (`AUTH_REQUIRED` defaults to false) and accepts any Host header. The 127.0.0.1 bind does not stop DNS rebinding. A web page can rebind its domain to 127.0.0.1, push an agent, and start a run with an image-based task it controls. It can also declare secret names, and `load_execution_secrets` copies those values from `secrets_file` into that container. Minimum fix: `TrustedHostMiddleware(allowed_hosts=["127.0.0.1", "localhost"])` in local mode. Better: generate a token at startup, write it to a 0600 file, and have the CLI read it as `api_key`.

services/tracker/main.py
```diff
@@ -634,14 +639,40 @@ async def _start_benchmark(
...
+                    "sandbox_provider": "docker",
+                    "properties": local_config.resources,
+                }
+            )
+        except ValueError as error:
+            raise HTTPException(status_code=400, detail=str(error)) from error
+        runtime = LocalRuntimeFactory.create_runtime(local_config.resources.data_root, run_starter.org.id)
+        aws_managed = False
+        effective_harness_config = None
+        managed_s3_bucket = None
```
BradenBug: blocking: A local Tracker sets `managed_s3_bucket` to `None` here instead of rejecting it. So `/start-benchmark-with-storage` against a local Tracker admits the run and executes it on local disk, and only then does the SDK raise `did not confirm requested storage bucket`. Please return 400 before admission when `managed_s3_bucket` is set, and add a test.

PR #810, opened 2026-09-17
Enable local agent library and authenticated SDK downloads

JarettForzano:
## Summary

Removed the environment-variable runtime switch, new agent download endpoint, and SDK relative-route handling after review. This implementation bypassed the planned configuration flow.

The local bundle-transfer changes now live in the runtime-adapter draft. This branch has no remaining tree difference from its base. Configuration-driven local agent operations remain unfinished.


docs/contributing/local-development.mdx
```diff
@@ -74,6 +74,11 @@ creates a disposable container, transfers a frozen agent bundle, executes it, ve
 its output, and deletes the container. Local transfers use provider file upload and
 preserve executable files without requiring curl or unzip in the image.
 
+Set `VALKYRIE_RUNTIME=local` and an absolute `VALKYRIE_LOCAL_DATA_ROOT` on Tracker
```
JarettForzano: we should not need env vars if we are using the config to define local sandbox provider.

services/tracker/src/tracker/api/agents.py
```diff
@@ -81,16 +85,35 @@ async def get_agent_download_url(
...
 
     return AgentDownloadURLResponse(name=name, download_url=url, expires_in=PRESIGNED_URL_EXPIRES_SECONDS)
 
 
+@router.get(
+    "/{name}/download",
+    response_class=Response,
+    responses={200: {"content": {"application/zip": {"schema": {"type": "string", "format": "binary"}}}}},
+)
+async def download_agent_archive(name: str, runtime: AgentLibraryRuntimeDependency) -> Response:
```
JarettForzano: i dont think we can add a new endpoint just for this

services/tracker/src/tracker/api/dependencies.py
```diff
@@ -79,12 +83,14 @@ async def get_run_runtime(run_context: RunAWSDependency) -> RuntimeServices:
 async def get_agent_library_runtime(
     request: Request,
     org: Org = Depends(get_current_org),
-) -> RuntimeServices:
+) -> AsyncGenerator[RuntimeServices]:
     """Open agent storage without constructing sandbox access."""
+    if os.environ.get("VALKYRIE_RUNTIME") == "local":
```
JarettForzano: no local env again

PR #819, opened 2026-09-18
Simplify async runtime operations before local adapters

JarettForzano:
## Summary

Make shared runtime operations async before #806 adds local adapters. Secret reads, Lambda calls, CloudWatch log group setup and ECS task protection await their clients directly.

Stack: dev → **#819** → #822 → #806 → #807 → #808.

## Changes

- Keep one async `SecretStore` interface. Remove the duplicate secret-read APIs, the sync Secrets Manager client, the `RuntimeServices` secret and provider wrappers, and `invoke_lambda`'s config default.
- Await Lambda invocation directly. The analyzer uses `total_max_attempts: 1`, so the non-idempotent analyzer Lambda is never retried. Dev's `max_attempts: 1` allowed one retry.
- The analyzer runs only its database status updates in worker threads, each with its own session.
- Read the sandbox cleanup secret on the sweep's event loop. The cleanup timeout now covers that read.
- Send ECS task protection updates with `httpx.AsyncClient` instead of `urllib` in a thread. ExecutorHost now pins `httpx==0.28.1`.
- Cache one async AWS session per event loop in `AWSClientProvider`.
- Drain executor claims and terminalization through repeated cancellation.
- Apply the retention policy when the log group already exists. Dev skipped it.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor
- [ ] Docs / config

## Testing

- [x] Manually tested

An AWS-mode run on the dev account with Daytona finished with score 100 at the #808 head. A forced stop deleted both sandboxes. Live AWS integration tests were not run.

## Checklist

- [x] Self-reviewed the diff
- [x] Docs updated if needed


BradenBug (approved):

BradenBug:
Love a net negative lines PR

PR #811, opened 2026-09-18
feat(tracker): classify task and run failures with FailureCategory

derrickvals:
### Why
Classifying errors clearly will set us up for more precise error handling and retry policies (vals-ai/valkyrie-ticket-library#178). 

### How
Added a FailureCategory describing the error as one of the following:
1. Infrastructure
2. Benchmark Service
3. Agent
4. Cancelled
5. Unknown

## Summary
Every task/run failure now carries a coarse, stable `FailureCategory` next to the existing `producer`/`operation`/`error_type`/`cause_code` provenance, decided at the `except` boundary where the failure is caught, and surfaced through the same structured field in the API, SDK, CLI, Sentry tags and structured logs.

Before this, read paths (single task, task list, final view, CLI `run errors`) only selected `error_message`, and the `tracker/process_task` catch-all reduced agent, sandbox and S3 failures to the same untyped bucket.

Retry eligibility and automatic retry are **not** touched (tracked in vals-ai/valkyrie-ticket-library#178); original-error preservation is vals-ai/valkyrie-ticket-library#282. Dashboards are a follow-up PR against `vals-ai/dashboards`. Run-level failures that only land on `Benchmark.error_message` (pre-task failures, all-tasks-failed summary) get no persisted category in this PR — that needs a `Benchmark` column/migration and is deferred.

## Failure category mapping

| Category | What it means | Who/what is at fault | Appropriate response |
|---|---|---|---|
| `infrastructure` | Valkyrie-owned plumbing failed: sandbox create/recover, artifact upload to S3, executor lease/claim expiry, enqueue, executor died without finishing | Vals infra (sandbox provider, AWS, executor fleet) | Operator: check provider/AWS health, then re-run the task. Retry-safe in principle — the agent never got a fair attempt. Automatic retry policy is #178. |
| `benchmark_service` | The per-task benchmark service (WebSocket) was unreachable, disconnected, returned malformed data, or rejected auth | Benchmark service / network between tracker and service | Operator: check benchmark-service deploy, DNS, auth credentials; re-run once healthy. Not an agent defect. |
| `agent` | The agent process itself exited non-zero / raised `AgentRunFailedError` inside a healthy sandbox | The agent under evaluation (or its harness contract) | Treat as a genuine agent failure for scoring purposes; inspect agent logs. Do **not** blindly retry — result is signal, not noise. |
[...]

PR #820, opened 2026-09-18
fix(tracker): preserve original agent and provider errors in task failures

derrickvals:
## Summary
Task failures currently lose the original error at the agent boundary: any non-zero `run_cmd`/`install_cmd` exit becomes `Agent command failed with exit code N` (PR #734 deliberately dropped the output tail to keep persisted errors prompt-free). This PR keeps that guarantee while letting agents opt in to a controlled error, labels install-phase failures, and appends the `__cause__` chain of wrapped provider errors to persisted task messages.

Stacked on #811 (`dl/failure-classification`); only this PR's commit is new.

## Changes
- `sandbox.stream_command_output` exports `VALKYRIE_ERROR_PATH=/tmp/.valkyrie/<run>.error` into the command's shell. On non-zero exit it reads at most 2048 chars from that file (`head -c`), whitespace-normalizes it, and appends it to the exit-code message; absent/unreadable file → unchanged exit-code-only message. The file is removed in the same `rm -f` as the timing files. Streamed stdout/stderr is still never persisted.
  ```
  Agent command failed with exit code 1
  Agent command failed with exit code 1: ValueError: model returned no patch
  ```
- Both post-exit reads (timing files, `$VALKYRIE_ERROR_PATH`) go through `_read_post_exit_file`: `asyncio.wait_for(_exec(...), timeout=5)` inside `try/except Exception` → `None`. The command's exit code is already known at that point, so sandbox loss, transport errors, or a hung read can no longer replace the exit-code `AgentRunFailedError` (or the success/timeout result); duration falls back to the tracker-side clock.
- New keyword-only `failure_subject` on `stream_command_output`; `_install_agent_dependencies_once` passes `Dependency installation for contract <name>` so install failures no longer read as agent-run failures.
- `task_execution._exception_message` walks `exc.__cause__` (stopping on cycles — `_process_task_attempt` re-raises the evaluation error with `raise e from e`, which made the first version of this walk loop forever and hang `tracker-unit-tests`) and appends `(caused by <Type>: <msg>)` for each cause whose text isn't already in the message, so `SandboxError(str(e)) from e` wrappers don't duplicate and `SandboxSetupError("Failed to clear egress rules") from TimeoutError(...)` surfaces the original.
- `docs/agents/agent-contract.mdx`: one paragraph + example documenting `$VALKYRIE_ERROR_PATH` under `run_cmd`.

Not changed: classification (`FailureCategory` from #811 is unaffected — the message grows, the type does not), retry policy (#178), cleanup paths (`rm -f` in `finally` and `_clear_egress_allowlist(fail_on_error=command_completed)` already protect the primary error; covered by existing tests).

## Follow-ups (not in this PR)
- Error-file content is agent-controlled: enforce/validate the documented `Type: message` shape and/or redact known-sensitive patterns before persisting, if the bounded opt-in channel is judged insufficient.

## Related Issues
Closes vals-ai/valkyrie-ticket-library#282

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor
- [x] Docs / config

## Testing
- [ ] Manually tested

Not live-tested: producing a real non-zero agent exit requires an executor, which the local compose stack does not run (same limitation as #811). The shell wiring (`export` → `timeout N sh -c '…'` writing `$VALKYRIE_ERROR_PATH` → `exit_code=$?` → `head -c 2048`) was exercised locally in `sh` and returned the written error with the original exit code, but not inside a provider sandbox.

## Checklist
- [x] Self-reviewed the diff
- [x] Docs updated if needed

Link to Devin session: https://app.devin.ai/sessions/05df1684a39246eaa73a0687982875b5
Open in Devin Desktop: https://app.devin.ai/desktop/session/05df1684a39246eaa73a0687982875b5?variant=devin
Requested by: @derrickvals

---

<a href="https://app.devin.ai/review/vals-ai/valkyrie/pull/820" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


derrickvals (commented):
Two items from an independent implementation of the same ticket (Codex, base #811 head), reconciled against head 23d62899.

**Required.** The reply on the `SandboxNotFoundError` thread says the read is deliberately left able to replace the agent failure, for consistency with the adjacent duration read. Ticket #282's done-when says the opposite: "A later collection or cleanup failure does not replace the original task failure." Reading the error file is exactly such a later step. `_read_agent_error` should catch every exception (not only `SandboxError`), bounded by a timeout, so the exit-code `AgentRunFailedError` always wins; the duration read should get the same treatment in this PR since it shares the boundary. Reference: the Codex branch wrapped both in `try/except Exception` with `asyncio.wait_for(..., timeout=5)` and kept the exit-code message on any failure.

**Should-fix.** No test follows a reported error from the sandbox boundary through storage to the task API. The unit tests stop at `stream_command_output` and `task_execution`. Add one producer test that writes an error file, exits non-zero, and asserts the `ErrorResult.error_message` (and, via #811, `category`) as read back through `get_single_task`.

**User choice, not a defect:** the file's content is accepted as free text after whitespace normalization. The Codex branch required the first line to match an exception-type pattern and dropped malformed files. Stricter, but it is a design call that belongs to the owner.

PR #822, opened 2026-09-19
Route agent publication and run bundle updates through Tracker

JarettForzano:
## Summary

Directory run starts and `resume/retry --update-agent` write to S3 from the CLI today. Route both through Tracker, so the server selects storage and applies org scoping.

Stack: dev → #819 → **#822** → #806 → #807 → #808.

**Breaking:** configs that set `DAYTONA_SECRET_NAME` without `sandbox_providers` stop working. Run `valkyrie config provider set <provider> <secret-name>`, then delete `DAYTONA_SECRET_NAME`. #808 adds this migration to `valkyrie config init`.

**Deploy order:** deploy Tracker before releasing this CLI. An older Tracker ignores `overwrite=false` and `update_agent=true`. It overwrites an existing alias and retries or resumes with the old bundle.

## Changes

- Add `overwrite=false` to agent upload. With it, Tracker publishes to S3 only if the alias is absent. An existing alias returns 409.
- Add `update_agent=true` to retry/resume. Tracker refreshes the saved bundle after authorization and task verification, before the recovery commit.
- Forward both options from the CLI and SDK. Remove the client-side S3 upload, copy and contract lookup.
- Remove the CLI's legacy `DAYTONA_SECRET_NAME` fallback and its docs.
- Regenerate the API and SDK references.

## Type of Change

- [x] Bug fix
- [x] New feature
- [x] Breaking change
- [x] Refactor
- [x] Docs / config

## Testing

- [x] Manually tested

Local Docker runs covered directory start without client AWS credentials, alias collision refusal and resume with an updated bundle. Runs at the #808 head finished with score 100 in local Docker mode and in AWS mode with Daytona. Only mocked unit tests cover S3 conditional-write collisions.

## Checklist

- [x] Self-reviewed the diff
- [x] Docs updated if needed


BradenBug (commented):
pls address Devin / Greptile comments

